PR #4, opened 2026-04-09
Scheduling AI Agent: Gmail push pipeline, LLM engine, and sidebar integration

### LINKED ISSUE #2: Scheduling AI Agent
# Context

The time has arrived! We have set the scaffolding in place, but now we can build the brain - the agent that decides what do to next.

Use our references to understand the workflow - we will add more details below.

# State transitions
This scheduling system is _almost_ like a finite state machine. Ideally, it is. However, the fact that all actions are humans emailing each other requires the agent to use some judgement and tact in terms of what to do next.

The happy path is pretty straightforward:
1. Client emails Coordinator asking to interview a Candidate
2. Coordinator forwards the email to Recruiter assigned to that Candidate
3. Recruiter confers with Candidate and sends Coordinator some times
4. Coordinator emails Client with list of times
5. Client replies with a specific time that works for both
6. Coordinator forwards to Recruiter so they know and can schedule a prep call, puts the interview on the calendar

However, these are emails between humans, so anything can happen.

For example: there might be no overlap between the interviewer availability and the candidate availability, requiring the coordinator to repeat steps 2-4

There may be multiple email threads for any recruiting loop

Sending emails should be techncially imposssible for the agent, as in its not one of its tools. 

## Task
Your job: design the AI agent system. Unless necessary, do not suggest new capabilities or API changes - try to use the API as it is.

You need a way to read the emails to do so. You need to do this via webhook - the coordinator should not have to open a message or the app for you to take action

The AI agent should suggest the next task to do, and that is what the recruiter should see. For example, if the next task should be "forward to recrutier", that is what she should see on her app sidebar.

nsadeh:
## Summary

Implements the Scheduling AI Agent RFC (#2) — a reactive layer that watches coordinator inboxes for scheduling-related emails, classifies them, drafts responses, and surfaces suggestions in the Gmail sidebar for coordinator approval.

- **Gmail Push Pipeline**: Pub/Sub webhook + arq background workers for near-real-time email processing with pull-based fallback (guaranteed delivery)
- **LLM Abstraction**: Anthropic primary / OpenAI fallback with two-model pipeline (Haiku for classification, Sonnet for drafting)
- **Agent Engine**: Email classification (8 categories), suggested actions (9 types), draft generation, guardrails (action-state validation, recipient validation)
- **Suggestion Persistence**: `agent_suggestions` + `suggestion_drafts` tables with full CRUD lifecycle
- **Sidebar Integration**: on-message shows agent suggestions with draft preview; 6 action handlers (approve, edit, reject, view, answer, create-loop)

### Architecture

```
Email arrives → Pub/Sub push → Webhook → arq job →
  Pre-filter (known thread/sender/keyword) →
  Agent Engine (classify → draft if needed) →
  Guardrails → Persist suggestion →
  Sidebar displays suggestion when coordinator opens email
```

### Graceful degradation

The agent degrades gracefully when services are unavailable:
- **No LLM API keys**: Pipeline runs but skips classification/drafting. Sidebar works in manual mode.
- **No Redis**: Webhook logs warning; push notifications not enqueued.
- **LLM provider down**: Automatic fallback to secondary provider; if both down, "Agent unavailable" banner.

### Follow-up work (Phase 5 of RFC)

- [ ] Langfuse tracing instrumentation
- [ ] Eval dataset seeding from MTTI archive
- [ ] Dashboard alerts (rejection rate, edit rate, latency)
- [ ] PII stripping input guardrail
- [ ] Load testing

## Test plan

- [x] 249 tests passing (5 pre-existing failures unrelated to this PR)
- [x] Pre-filter tests: keyword matching, known sender/thread, case insensitivity
- [x] Webhook tests: Pub/Sub parsing, coordinator validation, job enqueue, error handling
- [x] LLM tests: provider mocking, router fallback, error propagation
- [x] Agent engine tests: JSON extraction, classification parsing, draft generation, action skipping
- [x] Guardrail tests: action-state validation, create_loop rules, recipient validation
- [x] Worker tests: history sync, deduplication, watch renewal, debounce locks
- [x] Service tests: suggestion CRUD, draft attachment, resolve lifecycle
- [x] Lint clean (ruff check + format)

Closes #2


nsadeh:
# Code Review: Scheduling AI Agent

## Overall Assessment

Large, well-structured PR (~6700 additions, 32 files) implementing a reactive AI agent pipeline: Gmail push → pre-filter → classify → draft → coordinator approval. Architecture is sound. Findings organized by focus area.

---

## 1. Auth Regressions

### 🔴 HIGH: `_get_user_email` JWT not verified — `routes.py:60-86`

The function manually base64-decodes the JWT payload **without verifying the signature**:

```python
payload = parts[1]
payload += "=" * (4 - len(payload) % 4)
claims = json.loads(base64.urlsafe_b64decode(payload))
return claims.get("email")
```

The `userIdToken` is an OIDC token that identifies *which coordinator* is making the request. Without signature verification, a compromised session could impersonate another coordinator. The outer `verify_google_addon_token` dependency validates the *add-on request*, not the *user identity*.

**Fix:** Use `google.oauth2.id_token.verify_oauth2_token()` to validate against Google'\''s public keys.

### 🟡 MEDIUM: Migration 0005 orphans existing rows

```sql
ALTER TABLE agent_suggestions ADD COLUMN coordinator_email TEXT NOT NULL DEFAULT '\'''\''
```

Existing rows get `coordinator_email = '\'''\''`, which silently passes `NOT NULL` but breaks `WHERE coordinator_email = :coordinator_email` filters — those rows become invisible to all coordinators.

**Fix:** Backfill from the loop'\''s coordinator, or handle empty string in queries.

### 🟢 LOW: `_check_gmail_auth` accesses private `_token_store` — `routes.py:139`

```python
if not gmail._token_store:
```

Should expose a public method on `GmailClient` instead.

---

## 2. Functional Programming Design / State Bugs

### 🔴 HIGH: `_action_url` import bug — agent suggestion buttons are broken

**`agent/cards.py:44`** imports the raw module variable:
```python
from api.scheduling.cards import _action_url
```

Python strings are immutable — this captures the **initial value** (`""`) at import time, not the later-set value. The latest commit correctly changed `scheduling/cards.py` to use `get_action_url()`, but `agent/cards.py` still uses the raw import.

**Result:** All agent suggestion action buttons ("Send As-Is", "Edit", "Dismiss") will have `function=""` and silently fail when clicked.

**Fix:** `from api.scheduling.cards import get_action_url` and call `get_action_url()` in `_suggestion_action()`.

### 🔴 HIGH: Global mutable `_action_url` is a process-wide state leak — `scheduling/cards.py:45-53`

```python
_action_url: str = ""
def set_action_url(url: str) -> None:
    global _action_url
    _action_url = url
```

Called on every request via `_ensure_action_url()`. Under concurrent async requests, one request could set the URL, yield at an `await`, and another request could overwrite it before the first finishes building cards.

**Fix:** Use `contextvars.ContextVar` for per-request isolation, or pass the URL as a parameter to card builders.

### 🟡 MEDIUM: `send_email` drops multi-recipient drafts — `routes.py:713`

```python
to=draft.draft_to[0] if draft.draft_to else "",
```

Only the first recipient is passed. If the agent drafts to multiple recipients (e.g., confirmation to recruiter + client), extras are silently dropped. Also, empty `draft_to` passes `to=""` which may error or send to nobody.

**Fix:** Pass `to=draft.draft_to` (the full list).

### 🟢 LOW: Runtime imports inside `process_relevant_message`

`workers.py` lines 181, 187, 196 have runtime imports inside the function body. These add per-call latency and obscure the dependency graph. Move to module top if circular imports allow it.

---

## 3. Observability: Langfuse Tracing, Logging, Naming

### 🟡 MEDIUM: Double root traces per message

Both `workers.py` and `engine.py` create **top-level** Langfuse traces:
- `workers.py`: `langfuse.trace(name="process-message", ...)`
- `engine.py`: `langfuse.trace(name="agent-pipeline", ...)`

Each message generates **two unrelated traces** in the dashboard instead of a parent-child hierarchy.

**Fix:** Pass the worker'\''s `trace` into `run_agent()` and create spans under it, or remove the trace from `engine.py`.

### 🟢 LOW: Auto-instrumentation + manual spans = redundant data

`init_tracing()` patches the Anthropic SDK via `AnthropicInstrumentor`, which auto-creates spans for every `messages.create()`. Combined with manual `classify-email` / `draft-email` spans in `engine.py`, you get duplicate token/latency data. Consider whether you need both layers.

### ✅ POSITIVE: Logging is well-structured

- Consistent `logger.info` for business events, `logger.warning` for guardrails/config, `logger.debug` for skips
- Key-value format strings (`user=%s message_id=%s`) are grep-friendly
- Langfuse prompt names (`scheduling-classifier-system` etc.) follow clear conventions

---

## 4. General Code Quality, Bugs, Security

### 🔴 HIGH: Webhook endpoint has no authentication — `webhook.py`

`/webhook/gmail` is mounted without authentication:
```python
app.include_router(webhook_router)  # no Depends()
```

Anyone can POST a crafted Pub/Sub message to trigger processing of any coordinator'\''s mailbox. Google Pub/Sub supports push authentication via OIDC bearer tokens.

**Fix:** Add a `verify_pubsub_token` dependency, or verify the `Authorization` header against Google'\''s token endpoint.

### 🟡 MEDIUM: Pre-filter fetches full message after metadata — wasted API call

**`workers.py:292-303`:**
```python
metadata = await gmail.get_message_metadata(coordinator_email, msg_id)  # headers only
thread_id = metadata["threadId"]
message = await gmail.get_message(coordinator_email, msg_id)  # FULL message
```

The metadata call is wasted — the full message is fetched immediately after. Either adapt the pre-filter to work with metadata, or skip the metadata call.

### 🟡 MEDIUM: `Suggestion.__init__` uses fragile positional tuple indexing — `service.py:28-40`

```python
self.id: str = row[0]
...
self.coordinator_email: str = row[15] if len(row) > 15 else ""
```

If SQL column order changes, all indices break silently. The `len(row) > 15` guard confirms this is already causing backward-compat friction.

**Fix:** Use `psycopg`'\''s `row_factory=dict_row` or named tuples.

---

## Summary Table

| Severity | Count | Key Issues |
|----------|-------|------------|
| 🔴 HIGH | 4 | Unverified JWT, broken agent buttons (`_action_url` import), unauthenticated webhook, global state race |
| 🟡 MEDIUM | 4 | Migration orphaning, double Langfuse traces, dropped recipients, wasted API call, tuple indexing |
| 🟢 LOW | 4 | Private attr access, runtime imports, auto+manual spans, compose_email signature |

**Most urgent before merge:** The `_action_url` import bug in `agent/cards.py` — agent suggestion buttons will not work at all.

**Before production:** JWT verification and webhook authentication.

---
🤖 *Reviewed with Claude Code*

PR #8, opened 2026-04-13
Gmail push pipeline: real-time email processing

### LINKED ISSUE #6: gmail async capability
# Product Requirement Document Template

Feature: gmail async API integration

## Overview

To have a properly real-time scheduling agent, we need to react to email traffic in real time. Updating email threads when the user opens the email (as it is now) is insufficient for our proposed solution.

We need to be able to handle any incoming and outgoing message in the user's inbox with a handler. We also need to be able to see the message content/subject, sender, CC/BCC fields, target, and whether it's a forward, reply, or new thread.

## Goals

- Be able to handle any incoming or outgoing email in real time for all authenticated users in our user database
- Never miss an email, even when the service is down
- Associate emails to correct threads
- Access all important fields: subject, body, sender (if incoming), recipient (if outgoing), CC, BCC
- Identify whether the email is a forward, reply, or new email
- Make the reaction to these events accessible via an async hook

## Non-Goals

- Store any emails or email content
- Classification or AI content of any kind

## Audience

The audience is developers building user-facing features in this project. As such, it is more of a library/API than a user-facing feature.

## Existing solutions and issues

We can currently access message data by using the Workspace Addon API. However, this only works if the user opens a message with the add-on sidebar open.

This defeats the purpose of the application - the coordinator/user should not have to actually open all their emails. The agent should be able to work in real time. This adds substantial overhead and pain to the user experience.

## Assumptions

- Fewer than 100 emails
- Substantial email traffic - thousands per day
- Most scheduling actions will be a confirmation only, not requiring any further manual work from the user

## Constraints

We cannot drop emails, but we're allowed to be eventually consistent. This means that if the service is down, it should do a post-wakeup task to ensure it didn't miss anything.

## Key use cases

In the future, this will power workflows such as:
- Classifying emails as relating to workflows
- Advancing workflows through pre-defined rules/workbooks
- Drafting replies for emails
- Alerting individuals to certain things communicated via email

## Research

This is just the next step in the spec outlined in the references folder.

### User Research

N/A

#### Gmail Real-Time Notification Options

There are two viable approaches for reacting to Gmail mailbox changes. Both use the Gmail API; the difference is push vs. pull.

**Option A: Pub/Sub Watch + History Sync (Push-primary)**

- Call `users.watch()` per user to register a Cloud Pub/Sub subscription. Gmail pushes a lightweight notification (`emailAddress` + `historyId`) to our topic on every mailbox change.
- Notifications contain **no message content** — we must call `history.list()` and then `messages.get()` to fetch actual emails.
- Watches expire after **7 days**. Must renew proactively (e.g., every 6 hours via cron).
- Push is **best-effort**: notifications can be delayed, duplicated, or dropped. A polling fallback is required regardless.
- Requires: Google Cloud project with Pub/Sub enabled, topic with `gmail-api-push@system.gserviceaccount.com` as Publisher, OIDC token verification on webhook endpoint.
- Latency: ~2-10 seconds from email arrival to webhook hit (typical).
- Cost: `watch()` = 100 quota units/call. `history.list()` = 2 units. `messages.get()` = 5 units.
- Docs: https://developers.google.com/gmail/api/guides/push

**Option B: Periodic Polling (Pull-only)**

- Background worker calls `history.list()` per user on a fixed interval (e.g., every 1-2 minutes).
- Simpler operationally: no Pub/Sub topic, no webhook, no watch renewal, no OIDC verification.
- Latency: bounded by poll interval (1-2 min worst case).
- At our scale (~100 users, ~5,000 emails/day), this is ~50 history.list calls/min at 2 quota units each = 100 units/min, far below limits.
- Drawback: not "real-time" — 1-2 minute delay. But PRD says "eventually consistent" is acceptable.

**Option C: Google Workspace Events API**

- Newer subscription-based API for Google Workspace changes.
- As of early 2026, **Gmail is NOT a supported event source** — only Chat, Meet, and Spaces.
- Not viable for this use case.
- Docs: https://developers.google.com/workspace/events

**Recommendation:** Option B (polling) is dramatically simpler and sufficient at our scale. Option A adds significant operational complexity (Pub/Sub topic, webhook security, watch renewal) for ~1 minute of latency improvement. If latency requirements tighten later, we can add Pub/Sub push as an optimization on top of the poll-based foundation — the history.list() plumbing is identical.

#### History Tracking

- `history.list()` returns changes since a given `startHistoryId`. Each response includes a new `historyId` to use as the cursor for the next call.
- History IDs expire after **~30 days** of inactivity (returns HTTP 404). Recovery requires re-establishing a baseline via `messages.list()` or `getProfile()`.
- `historyTypes` filter: `messageAdded`, `messageDeleted`, `labelAdded`, `labelRemoved`. For our use case, `messageAdded` is the primary interest.
- Best practice: store `last_history_id` per user in the database. Advance it after each successful sync.

#### Reply vs. Forward vs. New Message Detection

- **Reply**: Has `In-Reply-To` and/or `References` headers. Gmail also threads by subject + participants, so `threadId` matching works. High reliability.
- **New thread**: No `In-Reply-To` / `References`, and `threadId` is new (not seen before). High reliability.
- **Forward**: **No standard header**. Gmail prepends "Fwd:" to subjects and quotes the original body, but this is a client convention, not a protocol guarantee. Outlook, Apple Mail, and others use different conventions. Detection requires heuristic parsing (subject prefix, quoted headers in body). Medium reliability at best.

#### Outgoing Email Tracking

- Gmail's `history.list()` with `messageAdded` captures both incoming and outgoing (Sent) messages.
- For outgoing messages our system sent: we already have the message ID from `drafts.send()` or `messages.send()`, so no additional tracking needed.
- For outgoing messages the coordinator sent manually: we'd see them via history sync, but we'd need to diff against "known" messages to identify coordinator-initiated sends. The `SENT` label and absence of our tracking metadata would distinguish these.

#### Authentication Model

- Our existing setup uses **per-user OAuth** with encrypted refresh tokens in Postgres (`gmail_tokens` table). This works for both push and pull approaches.
- `users.watch()` works with per-user OAuth — does not require domain-wide delegation.
- Domain-wide delegation would simplify token management (single service account, no per-user consent) but requires Google Workspace admin setup and is a larger scope change.

#### Idempotent Processing

- Gmail message IDs are globally unique and stable. Use them as deduplication keys.
- Options: Postgres table (`processed_messages`) or Redis SET with TTL.
- Redis is faster for hot-path dedup checks; Postgres is more durable. A reasonable hybrid: Redis for fast-path (TTL 24h), with the actual processing record in Postgres for audit.

#### Rate Limits (Gmail API)

| Limit | Value |
|-------|-------|
| Per-user burst | 250 quota units/second |
| Per-project daily | 1,000,000,000 units (default, adjustable) |
| `users.watch()` | 100 units/call |
| `history.list()` | 2 units/call |
| `messages.get(format=full)` | 5 units/call |
| `threads.get(format=full)` | 10 units/call |

At 100 users polled every 2 minutes: ~4,500 history.list calls/hour = 9,000 units/hour. Adding message fetches (~200 new messages/hour × 5 units) = 10,000 units/hour total. This is negligible.

#### Common Production Pitfalls

1. **Watch expiry silently stops push**: If renewal fails, no notifications arrive and no error surfaces. Polling fallback is the safety net.
2. **History gaps**: If historyId becomes stale (>30 days unused), `history.list()` returns 404. Must handle graceful re-baseline.
3. **Duplicate notifications**: Pub/Sub delivers at-least-once. Must deduplicate by message ID.
4. **Pub/Sub topic permissions**: `gmail-api-push@system.gserviceaccount.com` must have Publisher role — common setup mistake.
5. **Label filter mismatch**: `watch()` `labelIds` filter applies to messages, not changes. A message removed from INBOX still triggers if INBOX was watched.
6. **Thread grouping surprises**: Gmail threads by subject similarity, not just `In-Reply-To`. Unrelated emails with similar subjects may share a threadId.

nsadeh:
## Summary

Implements the Gmail async push pipeline RFC (#7, closes #6). This is a **decoupled infrastructure layer** in `api/gmail/` that watches coordinator inboxes via Gmail Pub/Sub push notifications and fires an async hook for downstream consumers (the scheduling agent).

- **Webhook**: OIDC-authenticated Pub/Sub endpoint (fixes PR #4's unauthenticated webhook bug)
- **Workers**: arq background jobs for push handling, 60s fallback poll, watch renewal, dedup cleanup
- **Hook interface**: `EmailHook` protocol with `LoggingHook` default — agent subscribes later
- **Classification**: Deterministic direction (incoming/outgoing) and type (new/reply/forward) via participant set diff
- **Scope validation**: Reads required scopes from `REQUIRED_SCOPES` env var, raises `GmailScopeError` on mismatch
- **Graceful degradation**: Works without Redis (poll-only) or Pub/Sub (manual watch registration)

### Files changed

| Area | Files | What |
|------|-------|------|
| Core pipeline | `hooks.py`, `webhook.py`, `workers.py` | Event model, OIDC webhook, arq workers |
| Gmail extensions | `client.py`, `auth.py`, `exceptions.py` | watch/history/profile methods, scope validation |
| Schema | `0003_gmail_push_pipeline.py`, `gmail_push.sql` | History tracking + dedup table |
| App wiring | `main.py` | Redis pool, webhook router, default hook |
| Dev scripts | `dev-worker.sh`, `dev-all.sh` | Worker process for local dev |
| Tests | `test_gmail_hooks.py`, `test_gmail_webhook.py`, `test_gmail_scope_validation.py` | 28 new tests |
| Docs | `references/gmail-push-setup.md` | Complete Pub/Sub + Railway setup guide |
| E2E test | `scripts/test_push_pipeline.py` | Self-send + cross-account incoming test |

### 1. Add Redis to Railway
- Railway project > **+ New** > **Database** > **Redis**
- Reference the Redis URL in the API service: `${{Redis.REDIS_URL}}`

### 2. Deploy the arq Worker
- Railway project > **+ New** > **Service** (same repo)
- Start command: `cd services/api && PYTHONPATH=src python -m arq api.gmail.workers.WorkerSettings`
- Add same env vars as API service (DATABASE_URL, REDIS_URL, GMAIL_TOKEN_ENCRYPTION_KEY, etc.)
- Worker does NOT need a public domain

### 3. Set Environment Variables
Add to both API and Worker services:

| Variable | Value |
|----------|-------|
| `PUBSUB_TOPIC` | `projects/ai-agents-dev-492713/topics/gmail-push` |
| `REDIS_URL` | `${{Redis.REDIS_URL}}` |
| `PUBSUB_WEBHOOK_AUDIENCE` | `https://api-staging-545f.up.railway.app/webhook/gmail` |
| `PUBSUB_SERVICE_ACCOUNT` | `pubsub-push@ai-agents-dev-492713.iam.gserviceaccount.com` |
| `REQUIRED_SCOPES` | `https://www.googleapis.com/auth/gmail.modify` |

### 4. Run Migration
```bash
railway run --service=api -- python -m yoyo apply --batch --database "$DATABASE_URL" ./migrations
```

### 5. Create Pub/Sub Subscription (if not done)
See `references/gmail-push-setup.md` for full instructions.

## Staging Test Plan

Run these tests in staging before promoting to production:

- [ ] **Migration**: Verify `processed_messages` table exists and `gmail_tokens` has `last_history_id`/`watch_expiry` columns
- [ ] **Worker startup**: Confirm arq worker logs show "Starting worker for 4 functions" and connects to Redis
- [ ] **Health check**: `curl https://api-staging-545f.up.railway.app/health` returns `{"status": "ok"}`
- [ ] **Webhook endpoint**: `curl -X POST https://api-staging-545f.up.railway.app/webhook/gmail` returns 200 (auth failure is logged, not returned)
- [ ] **Poll fallback**: Worker logs show `cron:poll_gmail_history` running every 60s without errors
- [ ] **Watch renewal**: Worker logs show `cron:renew_gmail_watches` running (every 6h, or trigger manually via test script)
- [ ] **E2E outgoing**: Run `test_push_pipeline.py --user nim@longridgepartners.com` — verify OUTGOING/NEW_THREAD classification
- [ ] **E2E incoming**: Send email from external account to coordinator — verify INCOMING/NEW_THREAD classification in worker logs
- [ ] **Forward detection**: Forward a thread email to a new recipient — verify FORWARD classification with correct `new_participants`
- [ ] **Reply detection**: Reply to an existing thread — verify REPLY classification
- [ ] **Dedup**: Send same email through both push and poll — verify only processed once (check `processed_messages` table)
- [ ] **Scope validation**: Temporarily add a fake scope to `REQUIRED_SCOPES` — verify `GmailScopeError` logged and coordinator skipped
- [ ] **Graceful degradation**: Stop Redis — verify API still starts (with warning) and poll runs without enqueueing
- [ ] **Existing tests**: Run `pytest tests/ --ignore=tests/test_scheduling_integration.py --ignore=tests/test_addon.py` — all 98 pass

### DIFF
--- a/services/api/queries/gmail_push.sql
+++ b/services/api/queries/gmail_push.sql
@@ -9,16 +9,17 @@ SET last_history_id = :last_history_id, updated_at = now()
 WHERE user_email = :user_email;
 
 -- name: update_watch_state(user_email, last_history_id, watch_expiry)!
--- Update both history cursor and watch expiration after watch registration.
+-- Update watch expiration and advance history cursor (never backward).
 UPDATE gmail_tokens
-SET last_history_id = :last_history_id,
+SET last_history_id = GREATEST(last_history_id, :last_history_id),
     watch_expiry = :watch_expiry,
     updated_at = now()
 WHERE user_email = :user_email;
 
 -- name: get_all_watched_emails
--- List all coordinator emails with stored tokens (for poll fallback).
-SELECT user_email FROM gmail_tokens;
+-- List coordinator emails with active watches (for poll fallback and watch renewal).
+SELECT user_email FROM gmail_tokens
+WHERE watch_expiry IS NOT NULL AND watch_expiry > now();
 
 -- name: is_message_processed(gmail_message_id)$
 -- Check if a message has already been processed (dedup).
--- a/services/api/queries/gmail_push.sql
+++ b/services/api/queries/gmail_push.sql
@@ -1,0 +1,37 @@
+-- name: get_history_id(user_email)$
+-- Load the last-processed Gmail history ID for incremental sync.
+SELECT last_history_id FROM gmail_tokens WHERE user_email = :user_email;
+
+-- name: update_history_id(user_email, last_history_id)!
+-- Advance the history cursor after successful sync.
+UPDATE gmail_tokens
+SET last_history_id = :last_history_id, updated_at = now()
+WHERE user_email = :user_email;
+
+-- name: update_watch_state(user_email, last_history_id, watch_expiry)!
+-- Update both history cursor and watch expiration after watch registration.
+UPDATE gmail_tokens
+SET last_history_id = :last_history_id,
+    watch_expiry = :watch_expiry,
+    updated_at = now()
+WHERE user_email = :user_email;
+
+-- name: get_all_watched_emails
+-- List all coordinator emails with stored tokens (for poll fallback).
+SELECT user_email FROM gmail_tokens;
+
+-- name: is_message_processed(gmail_message_id)$
+-- Check if a message has already been processed (dedup).
+SELECT EXISTS(
+    SELECT 1 FROM processed_messages WHERE gmail_message_id = :gmail_message_id
+) AS is_processed;
+
+-- name: mark_message_processed(gmail_message_id, coordinator_email)!
+-- Record a message as processed. ON CONFLICT handles race between push and poll.
+INSERT INTO processed_messages (gmail_message_id, coordinator_email)
+VALUES (:gmail_message_id, :coordinator_email)
+ON CONFLICT (gmail_message_id) DO NOTHING;
+
+-- name: cleanup_old_processed_messages!
+-- Delete dedup records older than 30 days.
+DELETE FROM processed_messages WHERE processed_at < now() - INTERVAL '30 days';
--- a/services/api/src/api/gmail/auth.py
+++ b/services/api/src/api/gmail/auth.py
@@ -8,12 +8,15 @@ from typing import TYPE_CHECKING
 from cryptography.fernet import Fernet, InvalidToken
 from google.oauth2.credentials import Credentials
 
-from api.gmail.exceptions import GmailAuthError, GmailUserNotAuthorizedError
+from api.gmail.exceptions import GmailAuthError, GmailScopeError, GmailUserNotAuthorizedError
 
 if TYPE_CHECKING:
+    from datetime import datetime
+
     from psycopg_pool import AsyncConnectionPool
 
-SCOPES = ["https://www.googleapis.com/auth/gmail.modify"]
+_DEFAULT_SCOPES = "https://www.googleapis.com/auth/gmail.modify"
+SCOPES = os.environ.get("REQUIRED_SCOPES", _DEFAULT_SCOPES).split(",")
 TOKEN_URI = "https://oauth2.googleapis.com/token"
 
 _LOAD_SQL = "SELECT refresh_token_encrypted, scopes FROM gmail_tokens WHERE user_email = %(email)s"
@@ -56,7 +59,11 @@ class TokenStore:
             )
 
     async def load_credentials(self, user_email: str) -> Credentials:
-        """Load a user's stored token and return Google OAuth Credentials."""
+        """Load a user's stored token and return Google OAuth Credentials.
+
+        Validates that stored scopes cover all REQUIRED_SCOPES. Raises
+        GmailScopeError if re-authorization is needed.
+        """
         async with self._pool.connection() as conn:
             cur = await conn.execute(_LOAD_SQL, {"email": user_email})
             row = await cur.fetchone()
@@ -64,6 +71,15 @@ class TokenStore:
         if row is None:
             raise GmailUserNotAuthorizedError(
                 f"No stored token for {user_email}. User must authorize via the add-on first."
+            )
+
+        granted = set(row[1])
+        required = set(SCOPES)
+        missing = required - granted
+        if missing:
+            raise GmailScopeError(
+                f"User {user_email} is missing scopes: {missing}. Re-authorization required.",
+                missing_scopes=list(missing),
             )
 
         refresh_token = self._decrypt(row[0])
@@ -93,3 +109,54 @@ class TokenStore:
             )
             row = await cur.fetchone()
             return row[0] if row else False
+
+    # --- Push pipeline state ---
+
+    async def get_history_id(self, user_email: str) -> str | None:
+        """Load the last-processed Gmail history ID for incremental sync."""
+        async with self._pool.connection() as conn:
+            cur = await conn.execute(
+                "SELECT last_history_id FROM gmail_tokens WHERE user_email = %(email)s",
+                {"email": user_email},
+            )
+            row = await cur.fetchone()
+            return row[0] if row else None
+
+    async def update_history_id(self, user_email: str, history_id: str) -> None:
+        """Advance the history cursor after successful sync."""
+        async with self._pool.connection() as conn:
+            await conn.execute(
+                """
+                UPDATE gmail_tokens
+                SET last_history_id = %(history_id)s, updated_at = now()
+                WHERE user_email = %(email)s
+                """,
+                {"email": user_email, "history_id": history_id},
+            )
+
+    async def update_watch_state(
+        self, user_email: str, history_id: str, watch_expiry: datetime
+    ) -> None:
+        """Update both history cursor and watch expiration after watch registration."""
+        async with self._pool.connection() as conn:
+            await conn.execute(
+                """
+                UPDATE gmail_tokens
+                SET last_history_id = %(history_id)s,
+                    watch_expiry = %(watch_expiry)s,
+                    updated_at = now()
+                WHERE user_email = %(email)s
+                """,
+                {
+                    "email": user_email,
+                    "history_id": history_id,
+                    "watch_expiry": watch_expiry,
+                },
+            )
+
+    async def get_all_watched_emails(self) -> list[str]:
+        """List all coordinator emails with stored tokens."""
+        async with self._pool.connection() as conn:
+            cur = await conn.execute("SELECT user_email FROM gmail_tokens")
+            rows = await cur.fetchall()
+            return [row[0] for row in rows]
--- a/services/api/src/api/gmail/webhook.py
+++ b/services/api/src/api/gmail/webhook.py
@@ -1,0 +1,117 @@
+"""Authenticated Gmail Pub/Sub webhook endpoint.
+
+Receives push notifications from Google Cloud Pub/Sub, verifies the
+OIDC bearer token, and enqueues an arq job for background processing.
+"""
+
+from __future__ import annotations
+
+import base64
+import json
+import logging
+import os
+
+from fastapi import APIRouter, Request, Response
+from google.auth.transport import requests as google_requests
+from google.oauth2 import id_token
+
+logger = logging.getLogger(__name__)
+
+webhook_router = APIRouter(tags=["gmail-push"])
+
+PUBSUB_SERVICE_ACCOUNT = os.environ.get(
+    "PUBSUB_SERVICE_ACCOUNT",
+    "gmail-api-push@system.gserviceaccount.com",
+)
+EXPECTED_AUDIENCE = os.environ.get("PUBSUB_WEBHOOK_AUDIENCE", "")
+if not EXPECTED_AUDIENCE:
+    logger.warning(
+        "PUBSUB_WEBHOOK_AUDIENCE not set — webhook OIDC audience "
+        "validation is disabled. Set this in production."
+    )
+
+
+async def _verify_pubsub_token(request: Request) -> dict:
+    """Verify the OIDC token Google attaches to Pub/Sub push messages.
+
+    Returns the verified claims dict, or raises ValueError on failure.
+    """
+    auth_header = request.headers.get("Authorization", "")
+    if not auth_header.startswith("Bearer "):
+        raise ValueError("Missing bearer token")
+
+    token = auth_header[7:]
+    claims = id_token.verify_oauth2_token(
+        token,
+        google_requests.Request(),
+        audience=EXPECTED_AUDIENCE or None,
+    )
+
+    if claims.get("email") != PUBSUB_SERVICE_ACCOUNT:
+        raise ValueError(f"Unexpected sender: {claims.get('email')}")
+
+    return claims
+
+
+@webhook_router.post("/webhook/gmail")
+async def gmail_webhook(request: Request) -> Response:
+    """Receive Gmail Pub/Sub push notifications.
+
+    1. Verify OIDC bearer token (Google-signed)
+    2. Parse emailAddress + historyId from Pub/Sub data
+    3. Validate coordinator has authorized the app
+    4. Enqueue arq job for background processing
+    5. Always return 200 (prevents Pub/Sub retries on errors)
+    """
+    # Verify OIDC token
+    try:
+        await _verify_pubsub_token(request)
+    except (ValueError, Exception) as exc:
+        logger.warning("webhook auth failed: %s", exc)
+        # Return 200 even on auth failure to prevent Pub/Sub retries
+        # from hammering the endpoint. Log for monitoring.
+        return Response(status_code=200)
+
+    # Parse Pub/Sub message
+    try:
+        body = await request.json()
+        message_data = body.get("message", {}).get("data", "")
+        decoded = base64.b64decode(message_data).decode("utf-8")
+        notification = json.loads(decoded)
+        coordinator_email = notification.get("emailAddress", "")
+        history_id = str(notification.get("historyId", ""))
+    except Exception:
+        logger.exception("webhook parse error")
+        return Response(status_code=200)
+
+    if not coordinator_email or not history_id:
+        logger.warning("webhook missing emailAddress or historyId")
+        return Response(status_code=200)
+
+    # Check if we have credentials for this coordinator
+    token_store = request.app.state.gmail._token_store
+    if not await token_store.has_token(coordinator_email):
+        logger.debug("webhook for unknown coordinator: %s", coordinator_email)
+        return Response(status_code=200)
+
+    # Enqueue background job
+    redis = getattr(request.app.state, "redis", None)
+    if redis is None:
+        logger.warning("redis not available — cannot enqueue push job")
+        return Response(status_code=200)
+
+    try:
+        await redis.enqueue_job(
+            "process_gmail_push",
+            coordinator_email,
+            history_id,
+        )
+        logger.info(
+            "enqueued push job coordinator=%s history_id=%s",
+            coordinator_email,
+            history_id,
+        )
+    except Exception:
+        logger.exception("failed to enqueue push job for %s", coordinator_email)
+
+    return Response(status_code=200)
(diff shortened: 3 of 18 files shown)


services/api/src/api/gmail/webhook.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+    auth_header = request.headers.get("Authorization", "")
+    if not auth_header.startswith("Bearer "):
+        raise ValueError("Missing bearer token")
+
+    token = auth_header[7:]
+    claims = id_token.verify_oauth2_token(
+        token,
+        google_requests.Request(),
+        audience=EXPECTED_AUDIENCE or None,
```
nsadeh: `audience=EXPECTED_AUDIENCE or None` passes `None` when the env var is unset (defaults to `""`). Per Google Auth library, `audience=None` skips audience validation entirely — any valid Google OIDC token is accepted. Should fail closed: raise or return 500 if `EXPECTED_AUDIENCE` is empty, rather than degrading to no-audience-check.

services/api/src/api/gmail/webhook.py
```diff
@@ -0,0 +1,117 @@
...
+    Returns the verified claims dict, or raises ValueError on failure.
+    """
+    auth_header = request.headers.get("Authorization", "")
+    if not auth_header.startswith("Bearer "):
+        raise ValueError("Missing bearer token")
+
+    token = auth_header[7:]
+    claims = id_token.verify_oauth2_token(
+        token,
+        google_requests.Request(),
```
nsadeh, line 54: `google_requests.Request()` makes a **synchronous** HTTP call to fetch Google's OIDC certs. This blocks the asyncio event loop. Use `asyncio.to_thread()` to wrap this, or use an async transport.

services/api/src/api/gmail/webhook.py
```diff
@@ -0,0 +1,117 @@
...
+    except Exception:
+        logger.exception("webhook parse error")
+        return Response(status_code=200)
+
+    if not coordinator_email or not history_id:
+        logger.warning("webhook missing emailAddress or historyId")
+        return Response(status_code=200)
+
+    # Check if we have credentials for this coordinator
+    token_store = request.app.state.gmail._token_store
```
nsadeh: Two issues on this line:

1. **Crash if `GMAIL_TOKEN_ENCRYPTION_KEY` is absent**: `app.state.gmail` is only set conditionally in `main.py`. This will raise `AttributeError` on every webhook hit if the key is missing. The existing `addon/routes.py` handles this gracefully with `getattr(request.app.state, "gmail", None)`.

2. **Private attribute access**: `_token_store` is a private attribute of `GmailClient`. PR #4 flagged this same pattern and suggested exposing a public method instead.

services/api/src/api/gmail/workers.py
```diff
@@ -0,0 +1,298 @@
...
+from __future__ import annotations
+
+import logging
+import os
+from datetime import UTC, datetime
+from pathlib import Path
+
+from dotenv import load_dotenv
+
+load_dotenv(Path(__file__).resolve().parent.parent.parent.parent / ".env")
```
nsadeh: `load_dotenv` at module import is a dev-only hack. In production (Railway), env vars are injected by the platform and `.env` doesn't exist. Project convention: "never write code that only works in dev; fix things the production way."

services/api/src/api/gmail/workers.py
```diff
@@ -0,0 +1,298 @@
...
+    gmail: GmailClient = ctx["gmail"]
+    emails = await token_store.get_all_watched_emails()
+
+    for coordinator_email in emails:
+        try:
+            result = await gmail.watch(coordinator_email, PUBSUB_TOPIC)
+            expiry = datetime.fromtimestamp(int(result["expiration"]) / 1000, tz=UTC)
+            await token_store.update_watch_state(
+                coordinator_email,
+                result["historyId"],
```
nsadeh, line 132: This unconditionally overwrites `last_history_id` with the watch response's `historyId`. If `_process_history` has already advanced the cursor past this point, the cursor moves backward, causing reprocessing on the next push/poll. The dedup table mitigates double-firing but doesn't prevent the race.

Fix: only update if the new ID is greater than the current one, or skip the history ID update during renewal.

services/api/queries/gmail_push.sql
```diff
@@ -0,0 +1,37 @@
...
+-- Update both history cursor and watch expiration after watch registration.
+UPDATE gmail_tokens
+SET last_history_id = :last_history_id,
+    watch_expiry = :watch_expiry,
+    updated_at = now()
+WHERE user_email = :user_email;
+
+-- name: get_all_watched_emails
+-- List all coordinator emails with stored tokens (for poll fallback).
+SELECT user_email FROM gmail_tokens;
```
nsadeh: This returns ALL coordinators with stored tokens, not just those with active Pub/Sub watches. `poll_gmail_history` runs every 60s against all of them, wasting Gmail API quota.

Suggested fix: `WHERE watch_expiry IS NOT NULL AND watch_expiry > now()`

nsadeh (commented):
### Code review

Found 8 issues:

1. **Auth bypass when `PUBSUB_WEBHOOK_AUDIENCE` is unset** — `audience=None` skips OIDC audience validation entirely, accepting any valid Google token.

https://github.com/nsadeh/lrp-scheduling-agent/blob/6034390e828e89a402190655eaeeb14d99ac9687/services/api/src/api/gmail/webhook.py#L26-L48

2. **Sync blocking HTTP call in async function** — `google_requests.Request()` makes a blocking HTTP request to fetch Google certs, stalling the event loop on every webhook hit.

https://github.com/nsadeh/lrp-scheduling-agent/blob/6034390e828e89a402190655eaeeb14d99ac9687/services/api/src/api/gmail/webhook.py#L43-L48

3. **Webhook crashes when `GMAIL_TOKEN_ENCRYPTION_KEY` is absent** — `request.app.state.gmail` is only set conditionally in `main.py`, but `webhook.py` accesses it unconditionally. Existing pattern in `addon/routes.py` uses `getattr(request.app.state, "gmail", None)`. Also accesses private `_token_store` — [PR #4 flagged this](https://github.com/nsadeh/lrp-scheduling-agent/pull/4).

https://github.com/nsadeh/lrp-scheduling-agent/blob/6034390e828e89a402190655eaeeb14d99ac9687/services/api/src/api/gmail/webhook.py#L91-L95

4. **`load_dotenv` at module import** — dev-only hack that won't work in Railway. Project memory explicitly says: "never write code that only works in dev; fix things the production way." ([CLAUDE.md](https://github.com/nsadeh/lrp-scheduling-agent/blob/6034390e828e89a402190655eaeeb14d99ac9687/CLAUDE.md))

https://github.com/nsadeh/lrp-scheduling-agent/blob/6034390e828e89a402190655eaeeb14d99ac9687/services/api/src/api/gmail/workers.py#L14-L17

5. **Watch renewal unconditionally overwrites history cursor** — `update_watch_state` resets `last_history_id` to the watch response's `historyId`, which may be older than the cursor `_process_history` has already advanced to. Races with in-progress processing can clobber a newer cursor.

https://github.com/nsadeh/lrp-scheduling-agent/blob/6034390e828e89a402190655eaeeb14d99ac9687/services/api/src/api/gmail/workers.py#L134-L139

6. **`get_all_watched_emails` polls ALL token holders** — returns every coordinator who ever OAuth'd, not just those with active watches. The 60-second poll wastes Gmail API quota on coordinators without active Pub/Sub watches.

https://github.com/nsadeh/lrp-scheduling-agent/blob/6034390e828e89a402190655eaeeb14d99ac9687/services/api/queries/gmail_push.sql#L19-L21

7. **New `GmailScopeError` not caught in `addon/routes.py`** — `load_credentials` now raises `GmailScopeError` for missing scopes, but `routes.py` only catches `GmailValidationError`. Scope errors will surface as generic 500s instead of prompting re-authorization.

https://github.com/nsadeh/lrp-scheduling-agent/blob/6034390e828e89a402190655eaeeb14d99ac9687/services/api/src/api/gmail/exceptions.py#L27-L33

8. **New env vars not documented in CLAUDE.md** — `PUBSUB_TOPIC`, `PUBSUB_WEBHOOK_AUDIENCE`, `PUBSUB_SERVICE_ACCOUNT`, `GMAIL_TOKEN_ENCRYPTION_KEY`, `REQUIRED_SCOPES` are all required but not listed in the [Environment Variables section](https://github.com/nsadeh/lrp-scheduling-agent/blob/6034390e828e89a402190655eaeeb14d99ac9687/CLAUDE.md#L40-L45). (CLAUDE.md `## Environment Variables`)

https://github.com/nsadeh/lrp-scheduling-agent/blob/6034390e828e89a402190655eaeeb14d99ac9687/CLAUDE.md#L40-L45

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

services/api/src/api/gmail/auth.py
```diff
@@ -93,3 +109,54 @@ async def has_token(self, user_email: str) -> bool:
...
+
+    async def update_watch_state(
+        self, user_email: str, history_id: str, watch_expiry: datetime
+    ) -> None:
+        """Update both history cursor and watch expiration after watch registration."""
+        async with self._pool.connection() as conn:
+            await conn.execute(
+                """
+                UPDATE gmail_tokens
+                SET last_history_id = %(history_id)s,
```
nsadeh, line 145: This inline SQL still does `SET last_history_id = %(history_id)s` without `GREATEST`. The fix was applied to `gmail_push.sql` but that file is dead code -- no loader connects it to the app. Apply the GREATEST fix here (with BIGINT cast, since the column is TEXT): `SET last_history_id = GREATEST(last_history_id::BIGINT, %(history_id)s::BIGINT)::TEXT`

services/api/src/api/gmail/auth.py
```diff
@@ -93,3 +109,54 @@ async def has_token(self, user_email: str) -> bool:
...
+                    "email": user_email,
+                    "history_id": history_id,
+                    "watch_expiry": watch_expiry,
+                },
+            )
+
+    async def get_all_watched_emails(self) -> list[str]:
+        """List all coordinator emails with stored tokens."""
+        async with self._pool.connection() as conn:
+            cur = await conn.execute("SELECT user_email FROM gmail_tokens")
```
nsadeh, line 160: This still returns ALL token holders. The active-watch filter (`WHERE watch_expiry IS NOT NULL AND watch_expiry > now()`) was added to `gmail_push.sql` but that file is never loaded. Apply the filter here.

services/api/queries/gmail_push.sql
```diff
@@ -0,0 +1,38 @@
...
+-- name: update_history_id(user_email, last_history_id)!
+-- Advance the history cursor after successful sync.
+UPDATE gmail_tokens
+SET last_history_id = :last_history_id, updated_at = now()
+WHERE user_email = :user_email;
+
+-- name: update_watch_state(user_email, last_history_id, watch_expiry)!
+-- Update watch expiration and advance history cursor (never backward).
+UPDATE gmail_tokens
+SET last_history_id = GREATEST(last_history_id, :last_history_id),
```
nsadeh, line 14: `GREATEST(last_history_id, :last_history_id)` does lexicographic comparison on TEXT columns. Gmail history IDs are numeric strings, so `GREATEST('9', '100')` = `'9'`. Use `GREATEST(last_history_id::BIGINT, :last_history_id::BIGINT)::TEXT` or change the column type to BIGINT.

nsadeh (commented):
### Code review (follow-up)

Of the 8 issues from the previous review, 6 were correctly fixed. 2 fixes were applied to dead code and the original bugs remain:

1. **Fixes #5 and #6 applied to `gmail_push.sql` but the Python code uses inline SQL in `auth.py`** -- `update_watch_state` (line 145) still does `SET last_history_id = %(history_id)s` without `GREATEST`, and `get_all_watched_emails` (line 160) still does `SELECT user_email FROM gmail_tokens` without the active-watch filter. The `queries/gmail_push.sql` file has both fixes, but no code loads it -- the scheduling module uses `aiosql` to load its SQL, but the gmail module has no equivalent loader. The fixes need to be applied to the inline SQL in `auth.py`.

https://github.com/nsadeh/lrp-scheduling-agent/blob/1ea9d71e0ca3e7d3cef5ab2e55d961388a6a22da/services/api/src/api/gmail/auth.py#L137-L162

2. **`GREATEST()` on TEXT column is lexicographic, not numeric** -- even once the fix is applied to `auth.py`, `GREATEST(last_history_id, :last_history_id)` does text comparison. Gmail history IDs are numeric strings, so `GREATEST('9', '100')` returns `'9'`. Cast to `BIGINT`: `GREATEST(last_history_id::BIGINT, :last_history_id::BIGINT)::TEXT`, or change the column type to `BIGINT` in the migration.

https://github.com/nsadeh/lrp-scheduling-agent/blob/1ea9d71e0ca3e7d3cef5ab2e55d961388a6a22da/services/api/queries/gmail_push.sql#L13-L17

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #14, opened 2026-04-14
feat(ai): Add AI infrastructure with LangFuse and LLM service

nsadeh:
### New Features
- **AI Module**: Created comprehensive AI infrastructure in `services/api/src/api/ai/`
  - `endpoint.py`: Typed FastAPI endpoints for AI operations
  - `llm_service.py`: Multi-provider LLM service with support for multiple AI providers
  - `langfuse_client.py`: LangFuse integration for observability and tracing
  - `errors.py`: Custom error handling for AI operations

### Testing
- Added comprehensive test suite:
  - `test_ai_endpoint.py`: Endpoint tests
  - `test_ai_llm_service.py`: LLM service tests
  - `test_ai_langfuse.py`: LangFuse integration tests
  - `test_ai_errors.py`: Error handling tests
  - `scripts/test_ai_integration.py`: Integration test script

### Dependencies
- Updated `pyproject.toml` with new AI-related dependencies
- Updated `uv.lock` with resolved dependency versions

### API Changes
- Integrated AI endpoints into main FastAPI application (`main.py`)
- Updated health check tests for API changes

## Summary
This PR establishes a complete AI infrastructure foundation with LLM service abstraction, observability through LangFuse, and comprehensive test coverage.

### DIFF
--- a/services/api/scripts/test_ai_integration.py
+++ b/services/api/scripts/test_ai_integration.py
@@ -34,10 +34,16 @@ from api.ai.llm_service import LLMService, init_llm_service  # noqa: E402
 
 # ── Test models ──────────────────────────────────────────────────────────────
 
-CLASSIFY_PROMPT = (
-    "Classify the following text into exactly one category: "
-    "greeting, question, statement. Text: {{text}}"
-)
+CLASSIFY_CHAT_PROMPT = [
+    {
+        "role": "system",
+        "content": (
+            "Classify the following text into exactly one category: "
+            "greeting, question, statement."
+        ),
+    },
+    {"role": "user", "content": "Text: {{text}}"},
+]
 
 
 class ClassifyInput(BaseModel):
@@ -73,7 +79,8 @@ async def test_langfuse_model_config(langfuse, llm):
     prompt_name = f"test-integration-model-config-{int(time.time())}"
     langfuse.create_prompt(
         name=prompt_name,
-        prompt=CLASSIFY_PROMPT,
+        prompt=CLASSIFY_CHAT_PROMPT,
+        type="chat",
         config={"model": "gpt-4o-mini", "temperature": 0.0, "max_tokens": 200},
         labels=["production"],
     )
@@ -107,7 +114,8 @@ async def test_default_model(langfuse, llm):
     prompt_name = f"test-integration-no-model-{int(time.time())}"
     langfuse.create_prompt(
         name=prompt_name,
-        prompt=CLASSIFY_PROMPT,
+        prompt=CLASSIFY_CHAT_PROMPT,
+        type="chat",
         config={},
         labels=["production"],
     )
@@ -152,7 +160,8 @@ async def test_failover(langfuse):
     prompt_name = f"test-integration-failover-{int(time.time())}"
     langfuse.create_prompt(
         name=prompt_name,
-        prompt=CLASSIFY_PROMPT,
+        prompt=CLASSIFY_CHAT_PROMPT,
+        type="chat",
         config={"model": "claude-sonnet-4-20250514", "temperature": 0.0, "max_tokens": 200},
         labels=["production"],
     )
@@ -252,7 +261,11 @@ async def test_traces_captured(langfuse, llm):
     prompt_name = f"test-integration-trace-{int(time.time())}"
     langfuse.create_prompt(
         name=prompt_name,
-        prompt="Reply with exactly: blue. Text: {{text}}",
+        prompt=[
+            {"role": "system", "content": "Reply with exactly: blue."},
+            {"role": "user", "content": "Text: {{text}}"},
+        ],
+        type="chat",
         config={"model": "gpt-4o-mini", "temperature": 0.0, "max_tokens": 50},
         labels=["production"],
     )
--- a/services/api/scripts/test_ai_integration.py
+++ b/services/api/scripts/test_ai_integration.py
@@ -1,0 +1,381 @@
+#!/usr/bin/env python3
+"""Integration test for the AI infrastructure.
+
+Runs against real LangFuse + LLM providers to verify:
+1. LangFuse prompt config drives model selection
+2. Defaults work when no model config in prompt
+3. Failover works when primary provider fails
+4. Traces are captured in LangFuse
+
+Usage:
+    cd services/api
+    PYTHONPATH=src uv run python3 scripts/test_ai_integration.py
+"""
+
+import asyncio
+import os
+import sys
+import time
+from pathlib import Path
+
+from dotenv import load_dotenv
+
+load_dotenv(Path(__file__).resolve().parent.parent / ".env", override=True)
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "src"))
+
+from pydantic import BaseModel  # noqa: E402
+
+from api.ai.endpoint import llm_endpoint  # noqa: E402
+from api.ai.errors import LLMUnavailableError  # noqa: E402
+from api.ai.langfuse_client import init_langfuse  # noqa: E402
+from api.ai.llm_service import LLMService, init_llm_service  # noqa: E402
+
+# ── Test models ──────────────────────────────────────────────────────────────
+
+
+class ClassifyInput(BaseModel):
+    text: str
+
+
+class ClassifyOutput(BaseModel):
+    category: str
+    confidence: float
+
+
+# ── Helpers ──────────────────────────────────────────────────────────────────
+
+PASS = "\033[92m✓\033[0m"
+FAIL = "\033[91m✗\033[0m"
+INFO = "\033[94mi\033[0m"
+
+
+def result(ok: bool, label: str, detail: str = ""):
+    mark = PASS if ok else FAIL
+    suffix = f" — {detail}" if detail else ""
+    print(f"  {mark} {label}{suffix}")
+    return ok
+
+
+# ── Test 1: LangFuse prompt config specifies model → that model is used ──────
+
+
+async def test_langfuse_model_config(langfuse, llm):
+    """Create a prompt in LangFuse with a specific model config and verify it's used."""
+    print("\n[Test 1] LangFuse prompt config drives model selection")
+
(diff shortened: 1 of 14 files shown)


services/api/src/api/ai/endpoint.py
```diff
@@ -0,0 +1,207 @@
...
+        input_type: type[BaseModel],
+        output_type: type[BaseModel],
+    ):
+        self.name = name
+        self.prompt_name = prompt_name
+        self.input_type = input_type
+        self.output_type = output_type
+        self._output_schema = json.dumps(output_type.model_json_schema(), indent=2)
+
+    @observe()
```
nsadeh: Does this observe call need to be instrumented with the prompt name to show up correctly on the trace? I think I did see a bunch of __call__ traces, is that intended?

services/api/src/api/ai/endpoint.py
```diff
@@ -0,0 +1,207 @@
...
+        temperature = overrides.get("temperature", config.get("temperature", 0.0))
+        max_tokens = overrides.get("max_tokens", config.get("max_tokens", 4096))
+
+        # 3. Compile the prompt template with input fields
+        input_dict = data.model_dump()
+        compiled_prompt = prompt.compile(**input_dict)
+
+        # 4. Build messages with JSON schema instruction
+        json_instruction = JSON_INSTRUCTION_TEMPLATE.format(schema=self._output_schema)
+        messages = [
```
nsadeh: There is usually a system prompt and a user prompt.

The user prompt is _usually_ a simple formatting of the input data, which is close to what you have here.

Is that enough? Or should we allow for an optional user prompt and fetch that from LangFuse?

services/api/src/api/ai/endpoint.py
```diff
@@ -0,0 +1,207 @@
...
+                self.name,
+                response.model,
+                response.provider,
+                response.latency_ms,
+            )
+            return parsed
+
+        # 7. Retry once with "fix your JSON" follow-up
+        logger.warning("Endpoint '%s': first parse failed, retrying with fix prompt", self.name)
+        fix_message = FIX_JSON_MESSAGE.format(
```
nsadeh, line 170: The whole JSON fixed loop is correct, I'm just wondering if guardrails.ai supports it natively? That seems up there alley

services/api/src/api/ai/langfuse_client.py
```diff
@@ -0,0 +1,103 @@
...
+    client = Langfuse(**kwargs)
+    logger.info("LangFuse client initialized (host=%s)", host or "cloud")
+    return client
+
+
+def fetch_prompt(
+    client: Langfuse,
+    name: str,
+    *,
+    label: str = "production",
```
nsadeh: Does LangFuse have the concept of prod vs a dev sandbox? If so, we should put the dev identifier into the .env so that we're able to iterate on prompts and see feedback from running the code locally without fucking up prod

services/api/src/api/ai/langfuse_client.py
```diff
@@ -0,0 +1,103 @@
...
+            f"Failed to fetch prompt '{name}' from LangFuse: {exc}"
+        ) from exc
+
+    if prompt.is_fallback:
+        logger.warning("Serving cached/fallback prompt for '%s' — LangFuse may be degraded", name)
+
+    return prompt
+
+
+class LangfuseFlushMiddleware(BaseHTTPMiddleware):
```
nsadeh: I always get wary when I see low level code like this when dealing with a 3rd party library - it often means we're using it weirdly. Can you verify that this is either a common pattern, or that we really need this for our use case?

services/api/src/api/ai/llm_service.py
```diff
@@ -0,0 +1,268 @@
...
+
+from api.ai.errors import LLMBudgetExceededError, LLMUnavailableError
+
+logger = logging.getLogger(__name__)
+
+# Silence litellm's verbose logging
+litellm.suppress_debug_info = True
+
+# Default model when LangFuse prompt config doesn't specify one
+DEFAULT_MODEL = "claude-sonnet-4-20250514"
```
nsadeh: Move to .env

services/api/src/api/ai/llm_service.py
```diff
@@ -0,0 +1,268 @@
...
+_PROVIDER_PREFIX_MAP = {
+    "anthropic/": "anthropic",
+    "openai/": "openai",
+    "gemini/": "google",
+}
+
+# Fallback model mapping: when the requested model's provider is down,
+# try equivalent-tier models on other providers. Each entry is an ordered
+# list of (provider_prefixed_model) to try.
+FALLBACK_MODEL_MAP: dict[str, list[str]] = {
```
nsadeh: Overly complicated, just have primary/secondary/tertiary in the .env.

services/api/scripts/test_ai_integration.py
```diff
@@ -0,0 +1,381 @@
...
+    await asyncio.sleep(1)
+
+    endpoint = llm_endpoint(
+        name="test_trace_capture",
+        prompt_name=prompt_name,
+        input_type=ClassifyInput,
+        output_type=ClassifyOutput,
+    )
+
+    await endpoint(
```
nsadeh: Does this actually call an LLM? Does any code in this test file? There should be at least one, ideally one from each of the three main model providers.

services/api/src/api/ai/endpoint.py
nsadeh: Good catch — not intended. The `@observe()` decorator was defaulting to the function name (`__call__`), which is useless in traces.

Fixed: we now call `langfuse.update_current_span(name=self.name)` inside `__call__` so traces show the endpoint name (e.g., `classify_email`).

services/api/src/api/ai/endpoint.py
nsadeh: What we have now is sufficient for the current use case (classification, drafting). The system prompt from LangFuse contains the task instructions, and the user message is the structured input data as JSON.

LangFuse does support "chat" type prompts (returns a list of messages instead of a single string), and our `fetch_prompt()` already accepts `prompt_type="chat"` — but the endpoint factory currently only handles text prompts. If we hit a case where we need separate system/user prompts that both need to be versioned independently, the simplest change would be to have the LangFuse prompt itself be a chat prompt with the system and user messages baked in. That keeps versioning in one place.

I'd defer adding the complexity until we actually need it for a specific endpoint. YAGNI for now — happy to revisit when we build the first endpoint that needs it.

services/api/src/api/ai/endpoint.py
nsadeh: Guardrails.ai does handle this — its `Guard` class wraps LLM calls and can validate + retry structured output against Pydantic models. It's literally designed for this.

However, it's a large dependency (pulls in its own LLM client, validator framework, telemetry). For what we need today — "parse JSON, retry once if it fails" — the 15 lines we have here are simpler and have no extra dependencies. The guardrails.ai approach would make more sense if we had complex validation rules (regex patterns, semantic checks, custom validators beyond Pydantic's built-in ones).

I'd say: keep the simple loop for now, and if we find ourselves adding validation logic beyond basic JSON schema conformance, migrate to guardrails.ai at that point. Worth tracking as a known option though.

services/api/src/api/ai/langfuse_client.py
nsadeh: Yes, LangFuse has two isolation mechanisms:

1. **`environment`** on the Langfuse client — tags all traces with a string like "development" or "production". Filters dev noise from prod in the dashboard.
2. **Prompt labels** — each prompt version can have labels ("production", "development", "latest"). `get_prompt()` fetches by label.

Fixed: added two new env vars:
- `LANGFUSE_ENVIRONMENT` — set to `"development"` in your `.env` to tag all local traces
- `LANGFUSE_PROMPT_LABEL` — set to `"development"` to pull dev prompt versions instead of prod

Both default to production behavior when not set.

services/api/src/api/ai/langfuse_client.py
nsadeh: You're right — this was unnecessary. LangFuse v3 uses OpenTelemetry under the hood with a `BatchSpanProcessor` that auto-flushes every 5 seconds. The per-request middleware was solving a problem that OTel's batching already handles.

Removed the `LangfuseFlushMiddleware` entirely. We still do a `langfuse.flush()` + `langfuse.shutdown()` in the lifespan teardown to ensure nothing is lost on process exit, which is the standard pattern from their docs.

services/api/src/api/ai/llm_service.py
nsadeh: Done — now reads from `LLM_DEFAULT_MODEL` env var, falls back to `claude-sonnet-4-20250514` if not set.

services/api/src/api/ai/llm_service.py
nsadeh: Done — replaced the `FALLBACK_MODEL_MAP` dict with two env vars:
- `LLM_SECONDARY_MODEL` (default: `gpt-4o`)
- `LLM_TERTIARY_MODEL` (default: `gemini-2.0-flash`)

The call chain is now just: primary (from LangFuse config or `LLM_DEFAULT_MODEL`) → secondary → tertiary. Deduplicates automatically if primary matches secondary.

services/api/scripts/test_ai_integration.py
nsadeh: This particular test (test 4/now test 5) did call an LLM — but you're right that there wasn't a test exercising each provider independently.

Fixed: added a new Test 4 "Direct provider calls" that creates a single-provider `LLMService` for each of Anthropic, OpenAI, and Gemini and makes a real call to each one. Results from latest run:
```
✓ anthropic: claude-haiku responded in 0.6s
✓ openai: gpt-4o-mini responded in 0.7s  
✗ google: gemini-2.0-flash — quota exceeded (free tier, 0 limit)
```
Gemini failure is a billing/quota issue with the current API key, not a code bug.

Issue #16, opened 2026-04-15
AI Drafts/Actions


nsadeh:
# Product Requirement Document Template

This PRD describes a desired system that allows the AI to do things, and how it applies to the first capability of the AI: drafts

## Overview

We have built the data layer, and we've built a classifier that decides what the appropriate next action should be. 

The next layer is to actually do stuff. In this iteration, we're only doing email drafts + sends.

However, in the future there will likely be other tasks, such as calendar updates and ATS updates, so we would like to design a system that can be extended to more task types. The choice to bundle the action system with a concrete action is deliberate, as to not engineer an abstraction that doesn't fit our use case.

These actions should be connected to the agent suggestions systems, in that when a user approves a suggestion, actions execute.

## Goals

1. When an agent suggestion is made that requires an email draft, one is created 
2. The draft follows specific instructions that differs based on who the target recipient is
3. When a user accepts the suggestion, the email draft is sent
4. It is easy to display the draft as part of the agent suggestion on a UI
5. It is easy to add additional task types and display their info on a UI in the future (extensible action system)

## Non-Goals

1. Actions before classification/suggestion generation are not supported
2. Changing the actions based on user disagreement with the suggested action is not supported
3. Scheduling or ATS updates will not be tackled in this PR
4. Actually rendering the suggestions/actions is not part of this feature, but we do want to make sure it's easy for a client to do by presenting a clean API for both updates and fetches

## Audience

The audience for the extensible action system is developers who are building actions for the agents

The audience for the drafts will be coordinators who approve a draft send from the UI (sidebar).

## Existing solutions and issues

Currently, users have to draft their own emails, which are formulaic and unnecessary.

## Assumptions

- Users are familiar with the scheduling process
- If a user disagrees with the suggestions/actions, they can send a custom email from the loop, where the classifier will automatically update the state of the loop and cancel invalid agent suggestions/actions
- Users must approve every outgoing email

## Constraints

- Users must approve every outgoing email. No AI agent has the ability to send an email 
- Users must be able to edit email drafts
- There's a specific template for displaying availability times (see in research below)

## Key use cases

All scheduling scenarios are part of our key use cases. This includes primarily messaging the clients - messages to recruiters are often empty forwards.

## Research

This section outlines any research questions that helped inform how this PRD was written. Although at the end, this section **must be completed before the PRD is written**, as it fully informs what the product is.

### User Research

1. What should be the tone of emails to clients?
The tone is professional but terse. Many one liners

[...]

PR #17, opened 2026-04-15
RFC: AI Email Draft Generation

nsadeh:
## Summary
- Adds RFC for the AI email draft generation pipeline (issue #16)
- Covers: data model (`email_drafts` table), draft generation pipeline (eager, separate LLM call), in-sidebar review/edit/send, recipient routing rules, feature flag rollout
- Resolves all open product questions from the #16 review thread

## Key Decisions Documented
- No polymorphic action framework — standalone `email_drafts` table with FK to `agent_suggestions`
- Eager generation (on classification, not on sidebar open)
- In-sidebar editing only (no Gmail native drafts)
- Hardcoded recipient routing by stage state
- Recruiter forwards skip LLM generation (empty body)
- Feature flag `DRAFT_GENERATION_ENABLED` for independent rollout

## Implementation Phases
1. Migration + model + DraftService (0.5d)
2. LangFuse prompt + drafter endpoint (1-2d)
3. ClassifierHook integration + feature flag (0.5d)
4. Sidebar cards + action handlers (1-1.5d)
5. Tests + eval dataset (1d)
6. Rollout + monitoring (1d)

Ref: #16

### DIFF
--- a/rfcs/rfc-ai-email-drafts.md
+++ b/rfcs/rfc-ai-email-drafts.md
@@ -1,0 +1,516 @@
+# RFC: AI Email Draft Generation
+
+| Field          | Value                                      |
+|----------------|--------------------------------------------|
+| **Author(s)**  | Kinematic Labs                             |
+| **Status**     | Draft                                      |
+| **Created**    | 2026-04-15                                 |
+| **Updated**    | 2026-04-15                                 |
+| **Reviewers**  | LRP Engineering, LRP Coordinator team      |
+| **Decider**    | Nadav Sadeh                                |
+| **Issue**      | #16                                        |
+
+## Context and Scope
+
+The email classification agent (RFC, PR #15) classifies incoming emails and persists structured suggestions to the `agent_suggestions` table — including `DRAFT_EMAIL` suggestions that indicate the coordinator should send an email. Today, `DRAFT_EMAIL` is a stubbed enum value: the classifier can suggest it, but nothing acts on it. Coordinators still manually compose every scheduling email, even though ~80% of them follow a small set of formulaic patterns (forward thread to recruiter, share availability with client, confirm interview time).
+
+This RFC proposes the **email draft generation pipeline**: when the classifier outputs a `DRAFT_EMAIL` suggestion, a second LLM call generates the email body, persists it to a new `email_drafts` table, and makes it available for coordinator review, editing, and one-click sending from the Gmail sidebar. Draft generation is eager (happens immediately on classification, not lazily when the sidebar opens) and all interaction happens in-sidebar.
+
+## Goals
+
+- **G1: Generate email drafts automatically when the classifier suggests one.** When a `DRAFT_EMAIL` suggestion is created, the system immediately generates a draft email body, persists it, and makes it available for the coordinator. No manual composition required for common scheduling patterns.
+- **G2: Match the tone and formatting coordinators use today.** Drafts must be professional, terse (often one-liners), and format availability in the specific `Candidate Name – Availability (ET):` format that clients expect. A coordinator should be able to send most drafts without editing.
+- **G3: Support in-sidebar review, editing, and sending.** The coordinator sees the draft in the Gmail sidebar, can edit the body inline, and sends with one click. The UX is: open sidebar → review → click send. No switching to Gmail compose, no copy-paste.
+- **G4: Maintain the human-in-the-loop constraint.** No email is ever sent without explicit coordinator approval. The draft pipeline generates and stores; only a coordinator action triggers the send.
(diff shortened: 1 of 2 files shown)


rfcs/rfc-ai-email-drafts.md
```diff
@@ -0,0 +1,516 @@
...
+- **REST API for drafts** — no new HTTP endpoints for listing, fetching, or managing drafts. The sidebar add-on is the only client. *Rationale:* the add-on calls directly into `LoopService` and `DraftService` via the action handler dispatch. A REST API can be added later against the same data model without restructuring.
+- **Recruiter-facing draft generation** — messages to recruiters are typically empty forwards of the client thread. The drafter does not generate body text for these; it sets `body=""` and the forward happens via existing `send_email` flow. *Rationale:* generating text for empty forwards adds LLM cost with no value. The coordinator's workflow for recruiter messages is already one click ("Forward to Recruiter").
```
nsadeh: Typically they're empty, but they can contain content. This RFC should support that. A recruiter can ask a question in a scheduling thread, the AI should try to respond to it if it has the information from the thread and loops.

rfcs/rfc-ai-email-drafts.md
```diff
@@ -0,0 +1,516 @@
...
+
+### What Coordinators Email Today
+
+Scheduling emails fall into a small number of patterns:
+
+| Pattern | Recipient | Typical Content | Frequency |
+|---------|-----------|----------------|-----------|
+| Request availability | Recruiter | Forward of client request, often with no body | Very high |
+| Share availability | Client | Candidate name + time slots in ET format | High |
+| Confirm interview | Client | Confirmation + logistics (phone number, Zoom link) | Medium |
```
nsadeh, line 45: Often clients send a zoom link to be used - the drafter should include in the message to the recruiter.

rfcs/rfc-ai-email-drafts.md
```diff
@@ -0,0 +1,516 @@
...
+```
+
+**Trade-off: eager vs. lazy generation.** We chose eager (generate immediately) over lazy (generate when sidebar opens). Eager wastes an LLM call if the coordinator rejects the suggestion, but these drafts are short and cheap (~200 tokens output). Lazy would add 1-3 seconds of latency to the sidebar open — unacceptable for the "open sidebar, click send" UX. At current classifier volume (~50 emails/day per coordinator, ~30% getting `DRAFT_EMAIL`), the wasted cost is negligible.
+
+#### 2. Recipient Routing
+
+Recipients are determined by stage state, matching existing logic in `_handle_compose_email` (addon/routes.py:396-424):
+
+| Stage State | Recipient | Subject Pattern |
+|-------------|-----------|-----------------|
```
nsadeh: We should probably keep the same subject for the entire thread?

Don't forget to CC the Client Manager on everything

It should be easily queryable who is who, because the suggestions may show it for clarity/confirmation

nsadeh (commented):
Review feedback — submitting to unblock replies.

rfcs/rfc-ai-email-drafts.md
nsadeh: Good catch — I was wrong to assume these are always empty. Updated the RFC:

- Removed the "recruiter-facing draft generation" non-goal. Recruiter drafts now go through the LLM like everything else, with `recipient_type="recruiter"` so the prompt can adapt.
- The drafter receives full thread context and generates a response when the classifier detects a question or information request. For simple forwards with no question, the LLM will produce a minimal/empty body naturally.
- Added "Answer recruiter question" as an email pattern in the Background table.
- Added test scenarios: "Recruiter forward (with question)" and "Recruiter forward (simple)".

rfcs/rfc-ai-email-drafts.md
nsadeh: Added. The classifier will extract Zoom links into `extracted_entities.zoom_link`, and the drafter prompt explicitly instructs: include it in confirmation/scheduling emails to the recruiter.

Also added a new email pattern row "Confirm interview (to recruiter)" and a test scenario "Confirmation to recruiter (with Zoom)" to cover this.

rfcs/rfc-ai-email-drafts.md
nsadeh: All three addressed:

**Same subject:** Drafts now always use the existing thread's subject (from `gmail_thread_id`). No more state-based subject generation like `Re: {title} - Availability Request`. Falls back to `Re: {loop.title}` only when there's no existing thread.

**CC client manager:** Added to the routing table — `loop.client_manager.email` is CC'd on every draft when present on the loop. This is now a hard rule in `_resolve_recipient()`, not optional.

**Queryable recipients:** Added a note on queryability — `email_drafts` stores `to_emails`/`cc_emails` as arrays, and the denormalized `loop_id` gives a single-join path to the full loop with all contacts. The sidebar can display "To: Haley Chen (Client) · CC: Mike Ross (CM)" with one join to loops + contacts. The CC open question is now marked resolved.

PR #18, opened 2026-04-16
feat: AI email draft generation pipeline

nsadeh:
## Summary

- Implements the AI email draft generation pipeline from RFC #17 / Issue #16
- When the classifier suggests `DRAFT_EMAIL`, a second LLM call generates the email body, persists it to `email_drafts`, and surfaces it in the Gmail sidebar
- Coordinators can review, edit inline, and send drafts with one click — matching the "open sidebar, click send" UX philosophy

### New module: `api/drafts/`
- **Migration 0005**: `email_drafts` table with FK to `agent_suggestions`, indexes on coordinator+status and suggestion_id
- **DraftService**: `generate_draft()` (LLM call with fallback to empty body), CRUD methods, recipient routing by stage state, LLM input construction
- **LLM endpoint**: `generate_draft_content` using `draft-email-v1` LangFuse prompt via the existing `llm_endpoint()` factory
- **Card builders**: `build_draft_preview()`, `build_draft_edit()`, `build_drafts_list_sections()`

### Modified files
- **classifier/hook.py**: After suggestion persistence, calls `DraftService.generate_draft()` when `action==DRAFT_EMAIL` and `linked_loop` exists
- **main.py + workers.py**: DraftService initialization gated on `DRAFT_GENERATION_ENABLED` env var (independent from `CLASSIFIER_ENABLED`)
- **addon/routes.py**: 4 new action handlers (`view_draft`, `edit_draft`, `send_draft`, `discard_draft`) + drafts tab shows pending AI drafts

### Feature flags
- `DRAFT_GENERATION_ENABLED=true` — enables draft generation. When false, `DRAFT_EMAIL` suggestions still created but no drafts generated
- Rollback: set to `false`, existing drafts stop appearing in sidebar

## Test Plan

- [x] 23 new tests pass (16 service unit tests + 7 card builder smoke tests)
- [x] 63 total tests pass (no regressions in existing classifier/scheduling tests)
- [x] ruff lint + format clean
- [ ] Create `draft-email-v1` prompt in LangFuse with few-shot examples
- [ ] Run migration against dev database
- [ ] Manual smoke test: trigger classification → verify draft row created → review/send from sidebar

## Next Steps

- Create the `draft-email-v1` LangFuse prompt with coordinator email examples and availability formatting rules
- Enable `DRAFT_GENERATION_ENABLED=true` for one coordinator, monitor LangFuse traces for 48h
- Measure draft acceptance rate (target: >60% sent without editing)

Ref: #16


services/api/src/api/drafts/models.py
```diff
@@ -0,0 +1,70 @@
...
+    created_at: datetime | None = None
+    updated_at: datetime | None = None
+
+
+# ---------------------------------------------------------------------------
+# LLM I/O models
+# ---------------------------------------------------------------------------
+
+
+class GenerateDraftInput(BaseModel):
```
nsadeh, line 52: I'm thinking it might be smart to keep the draft prompt as "dumb" as possible, almost like a tool.

That way, it's not necessary to put the whole state machine model in the prompt. The classifier, acting as the main agent brain that understands the workflow, gives the drafter an instruction of what to draft. The drafter's system prompt only contains tone instructions really.

To adjust the prompt to that, we probably need to adjust the input - less about the loop's state, more about the request to draft, and the thread summary should just be a formatted list of emails. If the draft directive is very tight, then the drafter doesn't need unbounded context of what's happening to not say something stupid.

services/api/src/api/drafts/service.py
```diff
@@ -0,0 +1,276 @@
...
+MAX_BODY_LENGTH = 2000
+
+
+async def _collect(async_gen) -> list:
+    return [row async for row in async_gen]
+
+
+def _row_to_draft(row: tuple) -> EmailDraft:
+    return EmailDraft(
+        id=row[0],
```
nsadeh: This seems really hacky to depend on indices like that, is this really how SQL reading in Python  works with Pydantic? Seems... wrong.

services/api/src/api/drafts/service.py
```diff
@@ -0,0 +1,276 @@
...
+        return loop.most_urgent_stage
+
+    def _resolve_recipients(
+        self,
+        loop: Loop,
+        stage: Stage | None,
+    ) -> tuple[list[str], list[str]]:
+        """Determine to/cc emails from stage state.
+
+        Mirrors the routing logic in addon/routes.py _handle_compose_email.
```
nsadeh: The logic should live in ONE place for ease of maintenance and DRY. Are you saying it's being moved here?

services/api/src/api/gmail/workers.py
```diff
@@ -64,6 +64,8 @@ async def startup(ctx: dict) -> None:
 
 def _init_classifier_hook(pool, gmail):
     """Create a ClassifierHook if AI infrastructure is available."""
+    import os
```
nsadeh: Update memory: all env vars live in a .env file or the env vars in Railway, use the dotenv package rather than os directly.

services/api/src/api/main.py
```diff
@@ -72,6 +72,21 @@ async def lifespan(app: FastAPI):
     app.state.langfuse = langfuse
     app.state.llm_service = llm_service
 
+    # Draft service — gated on separate feature flag
+    draft_service = None
+    draft_enabled = os.environ.get("DRAFT_GENERATION_ENABLED", "false").lower() == "true"
```
nsadeh: No feature flags!

services/api/src/api/drafts/models.py
nsadeh: Good call — redesigned in 4e8159b. `GenerateDraftInput` is now directive-driven:

- Removed `classification`, `stage_state`, `loop_context`, `recipient_type`
- Added `draft_directive` (populated from `suggestion.summary` — the classifier's natural-language instruction)
- `thread_messages` is a formatted email list for reply context
- `extracted_entities` remains as JSON for availability/phone/zoom data

The drafter prompt only needs tone rules + the directive. No state machine knowledge required.

services/api/src/api/drafts/service.py
nsadeh: You're right — the positional indexing is fragile. Fixed in 4e8159b by using psycopg's `dict_row` row factory:

```python
conn.row_factory = dict_row  # rows come back as dicts
row = await queries.get_draft(conn, id=draft_id)
return EmailDraft(**row)  # clean dict unpacking
```

This is a better pattern than the tuple-based approach used elsewhere in the codebase (classifier, scheduling). Could be adopted incrementally for the other services too.

services/api/src/api/drafts/service.py
nsadeh: Yes — centralized in 4e8159b. `resolve_recipients()` is now a module-level function in `drafts/service.py` (not an instance method). Both `DraftService.generate_draft()` and `_handle_compose_email` in routes.py import and call this single source of truth:

```python
from api.drafts.service import resolve_recipients
to_emails, _ = resolve_recipients(loop, stage)
```

One place to maintain, one place to test.

services/api/src/api/gmail/workers.py
nsadeh: The `import os` I added for `DRAFT_GENERATION_ENABLED` was removed in 0562a6f (feature flag removal). The pre-existing `os.environ` usage in workers.py for `DATABASE_URL`, `REDIS_URL`, etc. is an existing pattern — migrating that to dotenv is outside this PR's scope.

Updated memory with the env var convention.

services/api/src/api/main.py
nsadeh: Removed in 0562a6f — DraftService now initializes whenever LangFuse + LLM are available, matching the existing graceful degradation pattern. No env var check needed.

PR #37, opened 2026-04-20
Quote thread history when forwarding to new recipients (fix #36)

### LINKED ISSUE #36: Email context is not being forwarded
**Describe the bug**

When forwarding an email (e.g., to recruiter) by adding a new recipient, it does not sent the entire thread as expected in a true forward.

**To Reproduce**
1. Create a new loop
2. The classifier should output the first step - forward to recruiter
3. Forward to recruiter (you must control their mailbox)
4. Check the recruiter's mailbox. The email will appear there, but without the previous thread's context

**Expected behavior**
The recruiter should see the entire thread - otherwise they don't know what is happening

**Screenshots**
Image

See the screenshot - the recruiter has no idea what's going on, as the coordinator forwarded with an empty forward message (common) and the thread is not forwarded with.

**Additional context**
I can think of three potential fixes:

1. If the email send has forward APIs, use them when appropriate
2. Otherwise, a forward may just be a message with a formatted previous history
3. Another option is to not really forward, but use the AI to generate a concise but complete message, e.g., "please give us some times for candidate Claire Cao as she's interviewing for Millenium." The third goes against what the user currently does, but might actually provide a better user experiencer.

nsadeh:
## Summary

Closes #36. When the agent forwarded a scheduling thread to a recruiter, the recruiter received a bare note with no conversation context — Gmail only shows thread history to participants who were already on the thread, and the prior code only set `In-Reply-To` / `References` headers without quoting prior messages into the body.

**Fix:** build a Gmail-style quoted-history body server-side at send time, and prefix the subject with `Fwd:`. Applied only when `email_drafts.is_forward=True` (replies are untouched).

## What changed

- **New** `services/api/src/api/gmail/forward.py`
  - `build_forwarded_body(note, thread)` — appends chronological `---------- Forwarded message ----------` blocks below the coordinator's note.
  - `prefix_forward_subject(subject)` — idempotent, case-insensitive `Fwd:` prefix.
- **`services/api/src/api/addon/routes.py`** — `_handle_send_draft`:
  - Thread is fetched once and reused for both threading headers and forward-body construction.
  - On `is_forward=True`: body is rewritten with `build_forwarded_body` and subject gets `Fwd:`. The **persisted** draft body stays short — quoting only happens on the wire, so the sidebar view remains compact.
  - On `is_forward=True` with a failed thread fetch: raise. Silently sending a forward without context is the bug we're fixing; the pre-existing soft-fallback is kept for replies only.
- **Tests** (16 new, all passing):
  - `tests/test_gmail_forward.py` — formatting correctness, Cc omission, subject prefix idempotence.
  - `tests/test_addon_send_draft.py` — forward rewrites body/subject, reply path is unchanged, forward raises on fetch failure, reply still sends on fetch failure.

## Test plan

- [x] `uv run pytest` from `services/api` — **261 passed** (the 10 `test_ai_endpoint.py` failures are pre-existing Langfuse mock issues on `main`, unrelated to this PR).
- [x] `uv run ruff check` clean on changed files.
- [ ] **Manual end-to-end (recommended before merge):** run `./scripts/dev-all.sh`, create a loop that triggers the "forward to recruiter" suggestion, approve the draft, and confirm the recruiter's inbox shows the coordinator's note on top with the quoted `---------- Forwarded message ----------` history below and subject prefixed `Fwd:`.
- [ ] Regression manual check: send a non-forward reply and confirm body/subject are unchanged.

## Notes for reviewer

- `%-d` / `%-I` in `_format_message_block` are GNU/BSD strftime extensions — fine for macOS dev + Linux on Railway; would break on Windows.
- Subject prefix: matches Gmail's `Fwd:` dedupe. `Re:` is not stripped — a forwarded reply correctly becomes `Fwd: Re: ...`.
- Pre-existing failures in `tests/test_ai_endpoint.py` (Langfuse mock `AttributeError: name`) should be tracked separately.

### DIFF
--- a/services/api/src/api/addon/routes.py
+++ b/services/api/src/api/addon/routes.py
@@ -580,8 +580,14 @@ async def _handle_send_draft(body: AddonRequest, svc: LoopService, email: str, *
     if edited_body is not None and edited_body != draft.body:
         await draft_svc.update_draft_body(draft.id, send_body)
 
-    # Fetch the thread once — needed for threading headers and, on forwards, for
-    # quoting the prior history into the body.
+    # Fetch the thread once, up-front. Two downstream blocks consume it:
+    #   (1) the RFC 2822 threading headers immediately below, and
+    #   (2) build_forwarded_body() for forwards (issue #36).
+    # Keeping the fetch separate from its consumers also lets the error policy
+    # branch on is_forward: replies soft-fallback (warn + send with no
+    # threading headers — the recipient is already on the thread, so Gmail
+    # shows them context regardless); forwards raise, because a forward
+    # without quoted history is exactly the bug we're fixing.
     thread = None
     gmail = getattr(request.app.state, "gmail", None) if request else None
     if gmail and draft.gmail_thread_id:
@@ -589,9 +595,6 @@ async def _handle_send_draft(body: AddonRequest, svc: LoopService, email: str, *
             thread = await gmail.get_thread(email, draft.gmail_thread_id)
         except Exception:
             if draft.is_forward:
-                # A forward without quoted history leaves the new recipient with
-                # no context (issue #36). Fail loudly rather than silently send
-                # a bare note.
                 logger.exception(
                     "Failed to fetch thread %s for forward — aborting send",
                     draft.gmail_thread_id,
@@ -603,9 +606,15 @@ async def _handle_send_draft(body: AddonRequest, svc: LoopService, email: str, *
                 exc_info=True,
             )
 
-    # Resolve threading headers so the recipient sees this in the same thread.
-    # In-Reply-To + References are RFC 2822 headers that tell the recipient's
-    # email client to display this as a reply (or forward) in the thread.
+    # RFC 2822 threading headers so the recipient's Gmail stitches this message
+    # into the existing conversation rather than showing it as a new thread.
+    #   In-Reply-To: the Message-ID of the message we're directly replying to
+    #     (i.e. the last one in the thread).
+    #   References: a space-separated chain of every prior Message-ID in the
+    #     thread — lets Gmail place this message at the right node in the
+    #     thread tree even if some intermediate messages went missing.
+    # Without these, Gmail renders the outgoing message as a standalone
+    # conversation in the recipient's inbox.
     in_reply_to = None
     references = None
     if thread and thread.messages:
--- a/services/api/src/api/addon/routes.py
+++ b/services/api/src/api/addon/routes.py
@@ -22,6 +22,7 @@ from api.addon.models import (
 )
 from api.classifier.models import SuggestedAction, SuggestionStatus
 from api.gmail.exceptions import GmailScopeError, GmailValidationError
+from api.gmail.forward import build_forwarded_body, prefix_forward_subject
 from api.overview.cards import build_overview
 from api.overview.service import OverviewService
 from api.scheduling.cards import (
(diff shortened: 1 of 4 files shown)


services/api/src/api/addon/routes.py
```diff
@@ -579,36 +580,60 @@ async def _handle_send_draft(body: AddonRequest, svc: LoopService, email: str, *
...
-    references = None
+    # Fetch the thread once — needed for threading headers and, on forwards, for
+    # quoting the prior history into the body.
+    thread = None
     gmail = getattr(request.app.state, "gmail", None) if request else None
     if gmail and draft.gmail_thread_id:
         try:
             thread = await gmail.get_thread(email, draft.gmail_thread_id)
-            if thread.messages:
-                last_msg = thread.messages[-1]
```
nsadeh, line 592: Can you comment on why this change was necessary? What does it do?

services/api/src/api/addon/routes.py
```diff
@@ -579,36 +580,60 @@ async def _handle_send_draft(body: AddonRequest, svc: LoopService, email: str, *
...
                 draft.gmail_thread_id,
                 exc_info=True,
             )
 
+    # Resolve threading headers so the recipient sees this in the same thread.
+    # In-Reply-To + References are RFC 2822 headers that tell the recipient's
+    # email client to display this as a reply (or forward) in the thread.
+    in_reply_to = None
+    references = None
+    if thread and thread.messages:
```
nsadeh, line 620: What does this do?

nsadeh:
Answering both review questions:

---

### [routes.py:592](https://github.com/nsadeh/lrp-scheduling-agent/pull/37/files#diff-c793af44a5ac0beb8ab8a7ebd3e62df0c0c9d33af4a5a71efbe06b0e0aff4498R592) — _"Can you comment on why this change was necessary? What does it do?"_

Before this PR, the thread fetch and the `In-Reply-To` / `References` extraction lived in the **same** `try` block — the `Thread` object was used once inline and discarded. That was fine when threading headers were the only thing we needed.

Now the forward path **also** needs the `Thread` — to pass into `build_forwarded_body(send_body, thread)` for quoting prior messages into the body. Doing that inside the old nested try would tangle the error handling. So I hoisted the fetch:

1. Fetch once → bind `thread`
2. Reuse it for threading headers (existing behavior)
3. Reuse it for the forward-body rewrite (new)

The hoist also lets the error policy branch on `is_forward`:

- **Replies:** keep the existing soft-fallback — warn and send without threading headers (the recipient is already on the thread, so they have context natively).
- **Forwards:** `raise`. A forward without quoted history is exactly the bug in #36; silently sending one would re-introduce it. Fail loudly instead.

Same number of Gmail API calls as before — just one fetch whose result is now reused.

---

### [routes.py:611](https://github.com/nsadeh/lrp-scheduling-agent/pull/37/files#diff-c793af44a5ac0beb8ab8a7ebd3e62df0c0c9d33af4a5a71efbe06b0e0aff4498R611) — _"What does this do?"_

This is **unchanged** logic that was just moved out of the old try block (the diff makes it look new because the structure around it changed). It populates the RFC 2822 threading headers that Gmail uses to stitch a message into an existing conversation view:

- `In-Reply-To: <last-message-id>` → tells the recipient's Gmail "this message is a direct reply to that one."
- `References: <msg1-id> <msg2-id> …` → the full space-separated chain of every prior `Message-ID` in the thread, so Gmail can place this message at the right node in the thread tree even if some intermediate messages were missed.

Without these headers, Gmail would show the message as a standalone conversation in the recipient's inbox, breaking the scheduling thread. They're only populated if we successfully fetched the thread — otherwise we pass `None` and Gmail treats it as a new conversation (acceptable fallback for replies; forwards raise earlier so we never reach this point without a thread).

Happy to inline these explanations as code comments if you'd like — let me know.

services/api/src/api/addon/routes.py
nsadeh: Inlined the explanation as a code comment in 9be92f6 — TL;DR: the thread fetch was hoisted so **two** downstream blocks can reuse the same `Thread` object (the threading headers below, and `build_forwarded_body()` for forwards). Separating the fetch from its consumers also lets the error policy branch on `is_forward`: replies degrade softly, forwards raise (a context-less forward is the #36 bug).

services/api/src/api/addon/routes.py
nsadeh: Inlined a clearer explanation in 9be92f6. This block (unchanged logic, just relocated out of the old try) populates the RFC 2822 threading headers that Gmail uses to stitch the outgoing message into the existing conversation view in the recipient's inbox: `In-Reply-To` points at the last message in the thread, `References` is the full Message-ID chain of every prior message. Without them Gmail shows the message as a standalone thread.

PR #44, opened 2026-04-23
RFC: Inferred field prefill for manual create-loop workflow

### LINKED ISSUE #43: Infer details for create loop workflow
# Infer questions for create loop

For loops that the user creates manually (classifier misclassified or pre-classifier), all the fields are empty. This requires the user to enter all the details manually, which is tedious. We want to fill in whatever is possible for them.

## Overview

This PRD proposes filling the create loop UI from email details even through the manual workflow

## Goals

- When user clicks "create loop," an AI reads the email thread and parses out the form details
- A good, communicative loading state is presented while the AI works on that

## Non-Goals

N/A

## Audience

The audience is coordinators using the manual create loop workflow to create loops out of threads

## Existing solutions and issues

The current solution is to fill in the form manually. This is tedious and breaks our UX promise. It was also flagged by the primary customer POC.

## Assumptions

The same assumptions hold for other emails - external senders are the client company contact, CC'd long ridge employees are the client manager, the name of the candidate is in the email body and/or subject, it's impossible to figure out the recruiter from the email thread alone.

## Constraints

N/A

## Key use cases

N/A

### User Research

This request came directly in an email from the main POC of the customer. In the same email, his response to: "Is all this right so far?" was "From what I can tell we are far away."

### Technical Research

The simplest solution is to re-use the Pydantic model and run a lightweight extractor on the formatted thread.

nsadeh:
## Summary

- Proposes running a lightweight, typed LLM extractor against the full Gmail thread when the coordinator clicks "Create New Loop" on an unclassified thread, pre-filling candidate/client/company/CM/recruiter fields.
- Introduces a shared `CreateLoopExtraction` Pydantic model that both the classifier's `CREATE_LOOP` suggestions and the new manual-path extractor emit — replacing the current `dict[str, Any]` `extracted_entities` bag.
- Designs the loading state around a button-level spinner plus a post-render banner, with hard 5s timeout and graceful degradation to today's deterministic `from_` + `cc[0]` prefill on failure.

Closes #43.

## Design highlights

- **Scope limited to the manual path.** Classifier-path CREATE_LOOP suggestions keep using classifier output — this RFC does not rerun extraction for them.
- **Model choice: Haiku over Sonnet.** Narrower extraction task with strict schema enforcement; ~5x cheaper and ~2x faster than Sonnet to meet the ≤3s p95 latency goal.
- **Recruiter fields stay best-effort.** Prompt explicitly instructs the model to return null for recruiter rather than fabricate — sending the next draft to a wrong recruiter is worse than leaving the field blank.
- **Three-PR rollout:** (1) typed model + classifier migration, (2) extractor endpoint + LangFuse prompt + eval set, (3) wire into handler with loading state. Parallel-write to `extracted_entities` for one release.

## Test plan

- [ ] Reviewers confirm the scope split (manual only, classifier untouched) is the right call
- [ ] Reviewers weigh in on the loading-state choice (blocking spinner vs two-phase render — alternatives are in the RFC)
- [ ] Reviewers confirm Haiku is acceptable for this extraction (vs Sonnet on the classifier)
- [ ] Agreement on the rollout sequence and whether `extracted_entities` deprecation should happen in the same release or a later one

### DIFF
--- a/rfcs/rfc-infer-create-loop-fields.md
+++ b/rfcs/rfc-infer-create-loop-fields.md
@@ -1,0 +1,286 @@
+# RFC: Inferred Field Prefill for the Manual Create-Loop Workflow
+
+| Field          | Value                                      |
+|----------------|--------------------------------------------|
+| **Author(s)**  | Kinematic Labs                             |
+| **Status**     | Draft                                      |
+| **Created**    | 2026-04-23                                 |
+| **Updated**    | 2026-04-23                                 |
+| **Reviewers**  | LRP Engineering, LRP Coordinator team      |
+| **Decider**    | Nadav Sadeh                                |
+| **Issue**      | [#43](https://github.com/nsadeh/lrp-scheduling-agent/issues/43) |
+
+## Context and Scope
+
+When a coordinator clicks **"Create New Loop"** on a Gmail thread that the classifier did not flag as `CREATE_LOOP` — because the thread arrived before the classifier ran, the classifier misclassified it, or the coordinator is building a loop out of a thread that genuinely isn't a new interview request — the sidebar renders a nearly empty form. The only fields pre-filled are the `from_` address of the open message (as the client contact) and the first `cc` entry (as the client manager); everything else — candidate name, client company, recruiter name/email, first stage — is blank. See [`_handle_show_create_form`](services/api/src/api/addon/routes.py:541).
+
+This violates the product's core UX promise — "click, click, click, done". The primary customer POC flagged this directly: asked whether the experience so far matched expectations, his reply was *"From what I can tell we are far away."*
+
+Meanwhile, the classifier already does this kind of extraction for `CREATE_LOOP` suggestions — candidate name, client company, recruiter fields, etc. — and stores the output in a loose `extracted_entities: dict[str, Any]` on [`SuggestionItem`](services/api/src/api/classifier/models.py:74). The machinery works; we're just not running it on the manual path.
+
+This RFC proposes running a lightweight, typed LLM extractor against the full Gmail thread when the coordinator clicks "Create New Loop", pre-filling every field we can infer, and surfacing a clear spinner + message so the coordinator understands what the two-second wait is buying them.
+
+## Goals
+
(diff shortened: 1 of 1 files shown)


rfcs/rfc-infer-create-loop-fields.md
```diff
@@ -0,0 +1,286 @@
...
+Only the new manual-path extractor gets `CreateLoopExtraction`; the classifier keeps writing `dict[str, Any]`.
+
+**Upside:** smaller diff, no classifier-path test churn.
+**Downside:** we'd have the shared type already defined, the classifier already emits the same fields, and we'd ship two paths that differ only in schema enforcement. That's the worst of both worlds.
+
+**Decision:** reject. Typing once is cheaper than typing twice.
+
+## Risks and Open Questions
+
+- **R1: Latency overrun.** Haiku + short prompt should be well under 2s, but p99 excursions happen. **Mitigation:** set a hard 5s timeout on the extractor call (LiteLLM-level). If it exceeds, log and fall back to deterministic prefill — the coordinator sees the same form they see today, not an error. The spinner returns control in ≤5s no matter what.
```
nsadeh: 5s is too aggressive, make it 10

rfcs/rfc-infer-create-loop-fields.md
```diff
@@ -0,0 +1,286 @@
...
+- **R2: Hallucinated recruiter fields.** The prompt explicitly instructs "leave recruiter fields null unless clearly present," but models sometimes fabricate. **Mitigation:** add an eval set of ~15 threads (mix of recruiter-visible and recruiter-absent) and track recruiter-field false-positive rate specifically. Ship with a low threshold; tighten the prompt if fabrication shows up.
```
nsadeh: We should put a supervised eval here

rfcs/rfc-infer-create-loop-fields.md
nsadeh: Bumped to 10s in c93890b.

rfcs/rfc-infer-create-loop-fields.md
nsadeh: Promoted to a first-class deliverable of PR 2 in c93890b — labeled dataset of ~30 threads under `services/api/tests/evals/create_loop_extraction/`, per-field graders (with null-agreement and recruiter false-positive rate called out), LangFuse experiment wiring, and explicit ship-criteria thresholds that block PR 3 rollout.

PR #48, opened 2026-04-27
feat(classifier): pre-classifier sender blacklist for newsletters & notifications

nsadeh:
## Summary

Adds a YAML-driven sender blacklist that short-circuits the email classifier for senders that should never be clients — newsletters, LinkedIn notifications, Zoom voicemails, Adobe Sign requests, Eventbrite webinar promos, consumer/personal mail, and a few hand-curated cold-outreach addresses. When a blacklisted sender hits an unlinked thread, the LLM never runs, no `agent_suggestions` rows are written, and no Langfuse trace is emitted.

## Why

A Langfuse audit of today's production traces (`name=classify_email environment=production`) showed:

- **400** `classify_email` calls in 24 hours
- **338** (84.5%) classified `not_scheduling`
- **156** of those came from senders we'd never want the LLM to look at — newsletters (With Intelligence: 38, PitchBook: 17, etc.), LinkedIn auto-notifications, Zoom voicemail emails, etc.

The full audit is committed at `references/non-client-senders-2026-04-24.csv` so the seed list is reviewable end-to-end.

Each blacklistable email today costs:
1. Thread-history fetch + context assembly (`_build_context`)
2. A Claude call (~500 ms – 2 s, real $)
3. Up to 3 rows written to `agent_suggestions`
4. A Langfuse trace

This PR makes all four disappear for known non-client senders.

## How

- New YAML at `services/api/src/api/classifier/sender_blacklist.yaml` — 42 domains + 2 exact emails, organized by category with comments. Adding senders = open a PR.
- New loader at `services/api/src/api/classifier/sender_blacklist.py` — frozen dataclass, O(1) `is_blocked()` against a `frozenset` of lowercased domains and emails. Fail-open: missing/malformed file logs a warning and disables the blacklist rather than crashing the worker.
- `ClassifierHook` takes an optional `sender_blacklist` constructor param (defaults to empty) and checks it inside `on_email()` for **incoming emails on unlinked threads only**. Linked threads bypass the blacklist — if a newsletter ever shows up inside an active candidate conversation, the LLM still sees it.
- Wired into both production construction sites: the FastAPI lifespan (`main.py`) and the arq worker startup (`gmail/workers.py`).

## Design choices (asked the user)

- **Storage:** YAML file in repo (PR-reviewed, version-controlled) over Postgres table or env var.
- **Match granularity:** hybrid full-domain + exact-email. No suffix matching — `pitchbook.com` does NOT auto-match `news.pitchbook.com`. Each subdomain is listed explicitly to keep false-positive risk near zero.
- **Skip semantics:** silent. DEBUG log only; no DB write or Langfuse event. Aligns with the "open sidebar, click click, done" UX — these emails should disappear entirely.

## Tests

- `tests/test_sender_blacklist.py` (16 cases): domain match, exact match, case-insensitive, whitespace stripping, malformed YAML, missing file, wrong top-level type, empty file, end-to-end against the shipped YAML, malformed-address robustness.
- `tests/test_classifier_hook.py` (4 new in `TestSenderBlacklist`): blacklisted sender on unlinked thread skips, blacklisted sender on linked thread still classifies, non-blacklisted sender classifies normally, default empty blacklist when none injected.
- All 64 tests in the classifier-adjacent suites pass; ruff clean.

## Expected production impact

Tomorrow's Langfuse `classify_email` count should drop by roughly **120–150 calls/day** vs. today's baseline. The same Langfuse query is the audit tool — re-run it after deploy to confirm.

## What's deliberately out of scope

- A coordinator-facing UI to edit the blacklist. Future work if we ever want runtime curation.
- Suffix-domain matching. Adds false-positive risk; explicit subdomains for now.
- Auto-extending the blacklist from Langfuse traces (nightly cron). Worth revisiting after we see how often the seed list misses.
- Most of the `cold_outreach_spam` rows from the CSV. A few are genuine vendor relationships (e.g. Tricia Francis, the website vendor) that need human triage before adding. The CSV is the worklist.
- No `references/env-vars.md` or `references/railway-deployment.md` updates — no new env vars, no new Railway action; the YAML ships in the repo.

## Test plan

- [ ] Reviewer sanity-checks the seed YAML — anything in there look like a legit candidate/client domain?
- [ ] Reviewer scans `references/non-client-senders-2026-04-24.csv` `cold_outreach_spam` category for additional safe-to-block addresses worth adding to the YAML now.
- [ ] After merge + deploy, re-run the Langfuse query for the next 24 hours and confirm `not_scheduling` count drops by ~120–150.
- [ ] Spot-check `agent_suggestions` for the next few days — any blacklisted message_ids that should have produced a suggestion would indicate a false positive (most likely caught by reading `processed_messages` rows that have no matching `agent_suggestions`).


references/non-client-senders-2026-04-24.csv
```diff
@@ -0,0 +1,157 @@
+category,from,subject,timestamp,classification,trace_id,summary
```
nsadeh: this file is no longer required and may be deleted

services/api/sender_blacklist.yaml
```diff
@@ -0,0 +1,77 @@
+# Sender blacklist for the email classifier.
```
nsadeh, line 1: Is this the correct directory location for a config file?

references/non-client-senders-2026-04-24.csv
nsadeh: Done — deleted in 28aaa2c. The YAML header now points reviewers to the live Langfuse `classify_email` query as the audit source for new entries.

services/api/sender_blacklist.yaml
nsadeh: Good catch — moved to `services/api/sender_blacklist.yaml` (alongside `pyproject.toml`, `railway.toml`, `deployment.json`) in 28aaa2c. This matches the codebase convention where runtime data dirs (`queries/`, `migrations/`, `static/`) sit at the api root, not inside `src/`. The loader walks the same `parent x4` path that `api.scheduling.queries` uses to find `queries/`.

services/api/src/api/classifier/sender_blacklist.py
```diff
@@ -0,0 +1,92 @@
...
+from pathlib import Path
+from typing import Any
+
+import yaml
+
+logger = logging.getLogger(__name__)
+
+# services/api/src/api/classifier/sender_blacklist.py → services/api/sender_blacklist.yaml
+# Mirrors the path-walking pattern used by api.scheduling.queries (parent x4).
+DEFAULT_PATH = Path(__file__).resolve().parent.parent.parent.parent / "sender_blacklist.yaml"
```
nsadeh: Is this the cleanest way?

nsadeh: Good question — fair criticism. Replaced with a marker-file walk in 5909fdf:

```python
def _find_api_root() -> Path:
    for parent in Path(__file__).resolve().parents:
        if (parent / 'pyproject.toml').is_file():
            return parent
    raise RuntimeError(...)

DEFAULT_PATH = _find_api_root() / 'sender_blacklist.yaml'
```

This is robust to anyone moving `sender_blacklist.py` to a different depth inside `src/` — the chain of `.parent`s would silently break, but walking up to `pyproject.toml` works regardless. Same idiom as `python-dotenv`'s `find_dotenv`.

I deliberately didn't touch the three existing `api.*.queries` modules that use the same `.parent.parent.parent.parent` pattern — they're consistent with each other and a broader cleanup felt out of scope for this PR. Happy to do that as a follow-up if you'd like.

PR #49, opened 2026-04-28
Auto-resolve create_loop / advance_stage / link_thread; loops with incomplete contacts

nsadeh:
## Summary

- New `classifier/resolvers.py` registry auto-applies `create_loop`, `advance_stage`, and `link_thread` suggestions in the background — coordinators no longer see them as click-to-approve cards. Suggestions are marked `AUTO_APPLIED` so the overview filter (`WHERE status='pending'`) hides them.
- Loops can now be created with incomplete contact info — `loops.recruiter_id`, `loops.client_contact_id`, and `client_contacts.company` become nullable. Missing pieces are collected just-in-time on the draft card using the existing Workspace directory autocomplete.
- The Send button is disabled until a recipient is supplied; an inline candidate rename affordance shows when the auto-resolver fell back to "Unknown Candidate".

## Why

The first week of coordinator usage showed `create_loop` and `advance_stage` were friction, not value — the agent already knew the answer. Forcing recruiter/client emails up-front made `create_loop` impossible to auto-resolve since the recruiter address is rarely in the inbound interview request.

## Handoff plan

| Suggestion source | After deploy | Where it renders |
|---|---|---|
| Pre-deploy `PENDING` rows of these three types | Stay `PENDING` | Original card UI — coordinators clear backlog manually |
| New emails → resolver runs → `AUTO_APPLIED` | Filtered out by SQL | Never visible (happy path) |
| Resolver crashes (Sentry-and-drop) | Stays `PENDING` | Original card UI — manual completion fallback |

The original three card builders and `accept_suggestion` branches are intentionally kept so the backlog is finishable and resolver failures are recoverable. No data is dropped during the migration.

## Multi-loop threads

ClassifierHook now uses `find_loops_by_thread` (plural) and resolves `target_loop_id` per suggestion, so `create_loop` for a new loop B and `advance_stage` for an existing loop A can fire in the same classifier batch without conflict.

## Test plan

- [x] 13 new unit tests for resolvers (full extraction / empty extraction / partial / failure path / registry / try_auto_resolve)
- [x] Existing classifier-hook tests updated for `find_loops_by_thread`
- [x] Existing overview card tests preserved (the original three builders stay as fallback)
- [x] 307 unit tests pass; ruff clean
- [ ] Migration 0009 applies cleanly against a fresh dev DB (`docker compose up -d postgres && cd services/api && uv run yoyo apply --database \$DATABASE_URL`)
- [ ] End-to-end: send a fake interview-request email → verify no `create_loop` card appears → verify `DRAFT_EMAIL` card shows with inline recruiter inputs (when classifier didn't extract one) → pick recruiter → verify Send button enables → verify email goes out and recruiter is attached to the loop
- [ ] Backlog handoff: confirm any pre-deploy `PENDING` `create_loop` / `advance_stage` / `link_thread` suggestions still render with the old UI in staging
- [ ] Multi-loop thread: link a thread to two loops, classify a message → confirm `target_loop_id` routes the right action to the right loop

### DIFF
--- a/services/api/src/api/classifier/formatters.py
+++ b/services/api/src/api/classifier/formatters.py
@@ -100,7 +100,8 @@ def format_loop_state(loop: Loop | None) -> str:
     if loop.candidate:
         lines.append(f"Candidate: {loop.candidate.name}")
     if loop.client_contact:
-        lines.append(f"Client: {loop.client_contact.name} ({loop.client_contact.company})")
+        company = loop.client_contact.company or "Unknown"
+        lines.append(f"Client: {loop.client_contact.name} ({company})")
     if loop.recruiter:
         lines.append(f"Recruiter: {loop.recruiter.name} <{loop.recruiter.email}>")
 
@@ -114,6 +115,32 @@ def format_loop_state(loop: Loop | None) -> str:
     return "\n".join(lines)
 
 
+def format_linked_loops(loops: list[Loop]) -> str:
+    """Format ALL loops linked to the current thread.
+
+    Multi-loop threads (one Gmail thread linked to two or more loops, e.g.
+    two candidates discussed in the same chain) require the LLM to pick
+    which loop a suggestion targets via `target_loop_id`. This function
+    renders every linked loop so the LLM can disambiguate.
+    """
+    if not loops:
+        return "No matching loop found for this thread."
+
+    if len(loops) == 1:
+        return format_loop_state(loops[0])
+
+    blocks = [
+        f"This thread is linked to {len(loops)} loops. "
+        "When emitting a loop-scoped suggestion (DRAFT_EMAIL, ADVANCE_STAGE, "
+        "MARK_COLD), set `target_loop_id` to the specific loop you mean.",
+        "",
+    ]
+    for loop in loops:
+        blocks.append(format_loop_state(loop))
+        blocks.append("")
+    return "\n".join(blocks).rstrip()
+
+
 def format_active_loops(loops: list[Loop]) -> str:
     """Format coordinator's active loops summary for thread-to-loop matching."""
     if not loops:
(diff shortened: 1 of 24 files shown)


services/api/src/api/classifier/formatters.py
```diff
@@ -114,6 +115,32 @@ def format_loop_state(loop: Loop | None) -> str:
...
+    renders every linked loop so the LLM can disambiguate.
+    """
+    if not loops:
+        return "No matching loop found for this thread."
+
+    if len(loops) == 1:
+        return format_loop_state(loops[0])
+
+    blocks = [
+        f"This thread is linked to {len(loops)} loops. "
```
nsadeh: These instructions should live in the prompt, not the code.

services/api/src/api/gmail/workers.py
```diff
@@ -71,13 +71,20 @@ async def startup(ctx: dict) -> None:
         langfuse=langfuse,
     )
 
+    # Dedicated arq pool for the hook to enqueue follow-up reclassification
+    # jobs (after CREATE_LOOP / LINK_THREAD auto-resolve). Separate from arq's
+    # internal `ctx["redis"]` so it's lifecycle-managed by us.
+    arq_pool = await create_pool(RedisSettings.from_dsn(REDIS_URL))
```
nsadeh: Why do we need a separate redis pool? Why not just write back to the same one?

services/api/src/api/classifier/formatters.py
nsadeh: Good catch — moved out. The formatter now just renders the loop blocks; the multi-loop `target_loop_id` instruction needs to go in the LangFuse `scheduling-classifier-v3` prompt.

services/api/src/api/gmail/workers.py
nsadeh: You're right — dropped the second pool. Threading `ctx["redis"]` through `hook.on_email(event, arq_pool=...)` from each job, so we share arq's own pool now.

PR #58, opened 2026-04-30
fix(worker): eliminate Postgres connection pool starvation

nsadeh:
## Summary

The arq worker was hitting `TimeoutError` cascades on `reclassify_after_loop_creation` and other classifier jobs. Root cause: the connection pool was being starved by N+1 connection-acquisition patterns during email classification.

## What was happening

- `find_loops_by_thread` released its connection, then called `get_loop()` for each linked loop — and each `get_loop()` re-acquired a new connection to run ~7 nested queries (coordinator, contacts, candidate, stages, time slots, threads).
- `_get_active_loops` in the classifier hook reached directly into `LoopService._pool` and did the same N+1 acquisition.
- With `max_jobs=20` arq workers running concurrently against a 25-connection pool, an incoming push slammed the pool. Connection acquisitions blocked, jobs hit the 180s `job_timeout`, and the worker fell over.

## Fix

- **Decouple loop population from connection acquisition.** Extract `_populate_loop(conn, loop_id)` so callers that already hold a connection reuse it instead of acquiring a new one. Public `get_loop()` still acquires from the pool for external callers.
- **Single connection per batch.** `find_loop_by_thread`, `find_loops_by_thread`, `get_status_board`, and the new `get_loops_for_coordinator` now use one connection for the whole operation.
- **Replace `_get_active_loops`** with a call to the new public method (also removes the `_pool` reach-through code smell).
- **Bound pool waits** with `timeout=30.0` on both worker and API pools — starvation now surfaces as a fast `PoolTimeout` instead of blocking jobs until the 180s arq timeout.
- **Right-size concurrency.** Worker pool tightened to `min=5, max=20` and `max_jobs` reduced from 20 → 10, so peak concurrent jobs can't exceed pool capacity.

## Test plan

- [x] `pytest tests/test_classifier_hook.py` — 24 passed
- [x] `pytest tests/test_draft_service.py tests/test_classifier_resolvers.py tests/test_addon.py tests/test_overview_service.py` — 53 passed
- [x] Full suite (excluding pre-existing failures in `test_ai_endpoint.py`, `test_ai_langfuse.py`, `test_live_updates.py`, `test_scheduling_integration.py`) — 312 passed
- [ ] Watch worker logs after deploy for absence of `reclassify_after_loop_creation` `TimeoutError`s

## Reviewer notes

- `max_jobs=10` is intentionally conservative. With the N+1 fix, each job now uses 1–2 connections at a time, so this could likely be raised back toward 15 once we have telemetry on the new ceiling.
- No schema changes, no migration, no env var changes — drop-in fix.

### DIFF
--- a/services/api/src/api/classifier/hook.py
+++ b/services/api/src/api/classifier/hook.py
@@ -401,20 +401,7 @@ class ClassifierHook:
 
     async def _get_active_loops(self, coordinator_id: str) -> list[Loop]:
         """Load coordinator's active loops for thread matching context."""
-        from api.scheduling.queries import queries as sched_queries
-
-        async with self._loops._pool.connection() as conn:
-            rows = []
-            async for row in sched_queries.get_loops_for_coordinator(
-                conn, coordinator_id=coordinator_id
-            ):
-                rows.append(row)
-
-        loops = []
-        for row in rows:
-            loop = await self._loops.get_loop(row[0])
-            loops.append(loop)
-        return loops
+        return await self._loops.get_loops_for_coordinator(coordinator_id)
 
     def _apply_guardrails(
         self,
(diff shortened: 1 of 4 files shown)


nsadeh:
Closing — root cause analysis was off. Sentry breadcrumbs show aiohttp `Unclosed client session` warnings accumulating, pointing at a session leak from litellm under cancellation, not connection-pool starvation. The N+1 connection fix in this PR is real but not the cause of the timeouts. Replacing with a more targeted PR.

PR #61, opened 2026-05-01
perf(db): collapse N+1 Postgres queries into batch operations

nsadeh:
## Summary

Audit identified four hot paths where the API was issuing query counts proportional to the number of loops × stages — most painfully on the worker side, where every incoming email classification could fire 100+ queries just to build context. This PR collapses those into O(1) DB round-trips.

## Query counts (before → after)

| Hot path | Before | After |
|---|---|---|
| `get_loop()` (single loop) | 6 + S | **4** |
| `get_status_board()` (L loops, ΣSᵢ stages) | 2 + 6L + ΣSᵢ | **5** |
| `_get_active_loops()` (per-email, A active loops) | 1 + 6A + ΣSᵢ | **4** |
| `_process_history()` dedup (M messages) | 2M queries / 2M connections | **2 queries / 2 connections** |

Concrete examples:
- Status board with 30 loops × avg 2 stages: **242 → 5 queries**
- Email classification with 15 active loops × avg 2 stages: **121 → 4 queries per email**
- Push batch with 20 messages: **40 queries (40 conn checkouts) → 2 queries (2 conn checkouts)**

## Approach: multi-query fan-out, not mega-JOIN

Rather than a single multi-way JOIN (which causes row explosion across loops × stages × time_slots and is fragile against the existing positional tuple converters), the batch operations fan out to 3 flat queries — `get_stages_for_loops`, `get_time_slots_for_loops`, `get_threads_for_loops` — each using `WHERE id = ANY(:loop_ids)`. Python assembles the tree with `dict.setdefault` grouping. The query count is constant regardless of batch size.

## What changed

- **[services/api/queries/scheduling.sql](services/api/queries/scheduling.sql)** — Added `get_loop_full^` (single-loop JOIN over loop + coordinator + contacts + candidate), `get_loops_full_for_coordinator` and `get_active_loops_full_for_coordinator` (coordinator-batch JOINs), and `get_stages_for_loops` / `get_time_slots_for_loops` / `get_threads_for_loops` (batch by loop ID array).
- **[services/api/src/api/scheduling/service.py](services/api/src/api/scheduling/service.py)** — Added `_row_to_loop_full()` converter and `_get_loops_batch()` helper. Refactored `get_loop()` to use the JOIN query + the existing-but-unused `get_time_slots_for_loop`. Refactored `get_status_board()` to use the batch helper.
- **[services/api/src/api/classifier/hook.py](services/api/src/api/classifier/hook.py)** — `_get_active_loops()` now uses `get_active_loops_full_for_coordinator` + `_get_loops_batch`.
- **[services/api/src/api/gmail/workers.py](services/api/src/api/gmail/workers.py)** — Replaced per-message dedup loop with batch `SELECT ... ANY()` + batch `INSERT ... unnest() ... ON CONFLICT DO NOTHING`. At-most-once semantics preserved (all messages marked processed before any hook fires).

## Notes for reviewers

- The single-stage `get_time_slots_for_stage` query is kept (still used elsewhere), but `get_loop()` now calls `get_time_slots_for_loop` — that query already existed at scheduling.sql:319 but was never wired up.
- `_row_to_loop_full` uses positional tuple access (rows 0–31). The column layout is documented in its docstring; any change to the SQL must keep the order in lockstep.
- `get_active_loops_full_for_coordinator` uses `DISTINCT ON (l.id)` because the JOIN on `stages` would otherwise produce a row per active stage.
- The classifier hook adds a tiny local `_collect()` helper — same one as `scheduling/service.py`. Avoided importing a private function across modules.

## Test plan

- [x] `pytest tests/ --ignore=tests/test_scheduling_integration.py --ignore=tests/test_ai_endpoint.py` — 318 passed (test_ai_endpoint failure is pre-existing, unrelated mock issue; integration tests need a running Postgres)
- [x] `aiosql.from_path(...)` loads all 6 new queries without error
- [ ] **Reviewer to verify in dev**: `docker compose up -d` then `./scripts/dev-api.sh`, open the add-on sidebar with a coordinator that has multiple active loops, confirm status board renders correctly and check pg_stat_statements for query counts
- [ ] **Reviewer to verify in dev**: send a test email through the gmail push flow, confirm dedup works correctly when the same message_id arrives twice

### DIFF
--- a/services/api/queries/scheduling.sql
+++ b/services/api/queries/scheduling.sql
@@ -110,6 +110,72 @@ RETURNING id, coordinator_id, client_contact_id, recruiter_id, client_manager_id
 SELECT id, coordinator_id, client_contact_id, recruiter_id, client_manager_id, candidate_id, title, notes, created_at, updated_at
 FROM loops WHERE id = :id;
 
+-- name: get_loop_full^
+-- Fetch a loop with all actor details in a single query.
+SELECT
+    l.id, l.coordinator_id, l.client_contact_id, l.recruiter_id,
+    l.client_manager_id, l.candidate_id, l.title, l.notes,
+    l.created_at, l.updated_at,
+    -- coordinator (always present)
+    co.name, co.email, co.created_at,
+    -- client_contact (nullable)
+    cc.name, cc.email, cc.company, cc.created_at,
+    -- recruiter (nullable)
+    rec.name, rec.email, rec.role, rec.company, rec.photo_url, rec.created_at,
+    -- client_manager (nullable)
+    cm.name, cm.email, cm.role, cm.company, cm.photo_url, cm.created_at,
+    -- candidate (always present)
+    cand.name, cand.notes, cand.created_at
+FROM loops l
+JOIN coordinators co ON co.id = l.coordinator_id
+LEFT JOIN client_contacts cc ON cc.id = l.client_contact_id
+LEFT JOIN contacts rec ON rec.id = l.recruiter_id
+LEFT JOIN contacts cm ON cm.id = l.client_manager_id
+JOIN candidates cand ON cand.id = l.candidate_id
+WHERE l.id = :id;
+
+-- name: get_loops_full_for_coordinator
+-- All loops for a coordinator with actor details populated (for status board).
+SELECT
+    l.id, l.coordinator_id, l.client_contact_id, l.recruiter_id,
+    l.client_manager_id, l.candidate_id, l.title, l.notes,
+    l.created_at, l.updated_at,
+    co.name, co.email, co.created_at,
+    cc.name, cc.email, cc.company, cc.created_at,
+    rec.name, rec.email, rec.role, rec.company, rec.photo_url, rec.created_at,
+    cm.name, cm.email, cm.role, cm.company, cm.photo_url, cm.created_at,
+    cand.name, cand.notes, cand.created_at
+FROM loops l
+JOIN coordinators co ON co.id = l.coordinator_id
+LEFT JOIN client_contacts cc ON cc.id = l.client_contact_id
+LEFT JOIN contacts rec ON rec.id = l.recruiter_id
+LEFT JOIN contacts cm ON cm.id = l.client_manager_id
+JOIN candidates cand ON cand.id = l.candidate_id
+WHERE l.coordinator_id = :coordinator_id
+ORDER BY l.updated_at DESC;
+
+-- name: get_active_loops_full_for_coordinator
+-- Active loops (with >=1 non-terminal stage) for a coordinator, with actors.
+SELECT DISTINCT ON (l.id)
+    l.id, l.coordinator_id, l.client_contact_id, l.recruiter_id,
+    l.client_manager_id, l.candidate_id, l.title, l.notes,
+    l.created_at, l.updated_at,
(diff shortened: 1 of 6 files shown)


services/api/queries/scheduling.sql
```diff
@@ -110,6 +110,72 @@ RETURNING id, coordinator_id, client_contact_id, recruiter_id, client_manager_id
 SELECT id, coordinator_id, client_contact_id, recruiter_id, client_manager_id, candidate_id, title, notes, created_at, updated_at
 FROM loops WHERE id = :id;
 
+-- name: get_loop_full^
+-- Fetch a loop with all actor details in a single query.
+SELECT
```
nsadeh, line 122: Are you sure coordinator is _always_ populated? What happens if it isn't?

services/api/src/api/gmail/workers.py
```diff
@@ -273,34 +273,41 @@ async def _process_history(ctx: dict, coordinator_email: str, start_history_id:
             await token_store.update_history_id(coordinator_email, str(new_history_id))
         return
 
+    # Batch dedup: filter out already-processed messages in one query
+    async with pool.connection() as conn:
+        cur = await conn.execute(
+            """
+            SELECT gmail_message_id FROM processed_messages
+            WHERE gmail_message_id = ANY(%(ids)s)
```
nsadeh: Move this query to a sql file

services/api/src/api/gmail/workers.py
```diff
@@ -273,34 +273,41 @@ async def _process_history(ctx: dict, coordinator_email: str, start_history_id:
...
+        new_history_id = history_response.get("historyId")
+        if new_history_id:
+            await token_store.update_history_id(coordinator_email, str(new_history_id))
+        return
+
+    # Batch mark as processed BEFORE firing hooks (at-most-once)
+    async with pool.connection() as conn:
+        await conn.execute(
+            """
+            INSERT INTO processed_messages (gmail_message_id, coordinator_email)
```
nsadeh: Ditto

services/api/src/api/scheduling/service.py
```diff
@@ -282,53 +282,88 @@ async def create_loop(
...
             loop.stages = stages
 
-            # Populate email threads
+            # Threads (1 query)
             thread_rows = await _collect(queries.get_threads_for_loop(conn, loop_id=loop_id))
             loop.email_threads = [_row_to_email_thread(r) for r in thread_rows]
 
             return loop
 
+    async def _get_loops_batch(self, loop_rows: list[tuple]) -> list[Loop]:
```
nsadeh: Should loop rows by typed?

services/api/src/api/scheduling/service.py
```diff
@@ -796,6 +831,58 @@ def _row_to_loop(row: tuple) -> Loop:
...
+        updated_at=row[9],
+    )
+    loop.coordinator = Coordinator(id=row[1], name=row[10], email=row[11], created_at=row[12])
+    if row[13] is not None:
+        loop.client_contact = ClientContact(
+            id=row[2], name=row[13], email=row[14], company=row[15], created_at=row[16]
+        )
+    if row[17] is not None:
+        loop.recruiter = Contact(
+            id=row[3],
```
nsadeh: I don't like this index-based retrieval, can we make this more resilient?

PR #64, opened 2026-05-05
Collapse draft agent into next-action-agent v4

nsadeh:
## Summary

- **Removes the separate draft-email-v1 LLM call.** The next-action-agent prompt v4 now drafts the email body directly in `action_data.body`, eliminating the two-agent (classifier → drafter) workflow.
- **Adds actor name template variables** (`candidate_name`, `recruiter_name`, `client_name`, `client_company`) to `NextActionInput` for the v4 prompt.
- **Fixes forward detection** to check only the triggering message's participants rather than the entire thread history — a forward is when the recipient wasn't on the *last* message, not when they've never appeared in the thread.

## Details

**Schema changes:**
- `DraftEmailData.directive` → `DraftEmailData.body`
- `NextActionInput` gains 4 new fields extracted from the primary linked loop

**Dead code removed:**
- `services/api/src/api/drafts/endpoint.py` (entire file)
- `GenerateDraftInput` and `DraftOutput` models
- `DraftService._build_input()`, `_resolve_recipient_name()`
- `llm`/`langfuse` params from `DraftService.__init__` and all constructor call sites

**No DB migration needed.** Old PENDING suggestions with `action_data.directive` already have `EmailDraft` rows with generated bodies — the sidebar reads `draft.body`, not `action_data`.

**Deploy order:** Push v4 prompt to LangFuse first, then deploy this code.

## Test plan

- [x] All unit tests pass (286 passed)
- [ ] Manual: trigger email on linked thread → verify agent produces `action_data.body` → verify draft row created → verify sidebar shows editable body → send works
- [ ] Verify existing PENDING suggestions still render and send correctly
- [ ] Verify forward detection: drafting to a recipient not on the triggering message shows "Forward" button


services/api/src/api/classifier/next_action_agent.py
```diff
@@ -254,16 +256,32 @@ async def _build_context(
...
         coord = await self._loops.get_coordinator_by_email(event.coordinator_email)
         coordinator_name = _resolve_coordinator_name(event, coord)
-        coordinator_str = f"{coordinator_name}<{event.coordinator_email}>"
         date_str = datetime.now(UTC).date().isoformat()
 
         events = []
         if linked_loops:
             events = await self._loops.get_events(linked_loops[0].id)
 
+        primary = linked_loops[0] if linked_loops else None
```
nsadeh: This makes no sense. 

An email on a thread can be about _any_ loop attached to that thread. Taking the first one is completely random.

We should have a different fallback state or we need a different way to pass this in. Maybe a formatted list like this

- lop_1234: Re CruiterA
- lop_2345: Re CruiterB

And so on for the candidates as well

An email thread _will_ only ever have one client, though, the client contact may be different for different loops.

nsadeh: Fixed in 6f0f2a3. 

`candidate_name` and `recruiter_name` are now formatted per-loop:

- **Single loop** → plain name (e.g., `Sarah Chen`)
- **Multi-loop** → list with loop IDs:
  ```
  - lop_abc: Sarah Chen
  - lop_def: Mike Johnson
  ```

`client_name` and `client_company` take the first non-null value since they're shared across loops on a thread.

PR #68, opened 2026-05-06
feat(agent): deduplicate suggestions via pending_suggestions context

nsadeh:
## Summary

- Adds a `pending_suggestions` template variable to the next-action-agent prompt, populated with all currently pending suggestions for loops on the thread. This gives the LLM visibility into what's already queued so it can avoid proposing duplicates.
- Adds a hard dedup check: if a new suggestion matches an existing pending suggestion on (loop_id, action, action_data), it's discarded. Also catches same-batch duplicates within a single LLM response.
- New `format_pending_suggestions()` formatter renders pending suggestions in a human-readable format for the prompt.

## LangFuse action needed

After merging, add `{{pending_suggestions}}` to the "next-action-agent" prompt template with guidance to avoid duplicating existing suggestions.

## Test plan

- [x] `test_duplicate_of_existing_pending_is_skipped` — verifies exact match against DB is discarded
- [x] `test_same_batch_duplicate_only_creates_first` — verifies LLM returning 2 identical items only persists one
- [x] `test_different_action_data_not_deduplicated` — verifies different action_data is NOT falsely deduplicated
- [x] `test_empty_returns_default` / `test_multiple_suggestions_formatted` — formatter coverage
- [x] All 50 existing + new tests pass

### DIFF
--- a/services/api/src/api/classifier/formatters.py
+++ b/services/api/src/api/classifier/formatters.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
+    from api.classifier.models import Suggestion
     from api.gmail.models import Message
     from api.scheduling.models import Loop, LoopEvent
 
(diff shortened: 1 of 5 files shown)


services/api/src/api/classifier/formatters.py
```diff
@@ -179,3 +180,35 @@ def format_stage_states() -> str:
...
+                line += f" ({highlights})"
+        lines.append(line)
+
+    return "\n".join(lines)
+
+
+def _action_data_highlights(action: str, action_data: dict) -> str:
+    if action == "advance_stage":
+        return f"target_stage={action_data.get('target_stage', '?')}"
+    elif action == "draft_email":
```
nsadeh, line 207: Need body too

PR #76, opened 2026-05-08
fix: prevent classifier mislinks and loop explosion

nsadeh:
## Summary

A production LangFuse trace revealed the loop classifier linking an email about "Wongmo Kang, Marshall Wace" to the wrong loop (Evan Fiedler's Marshall Wace loop). Root cause: thousands of duplicate loops overloaded the LLM context, causing confused classification.

Fixes at four layers:

- **Data cleanup** — migration `0013` marks duplicate `state='new'` loops as `cold`; `create_loop()` now dedup-checks by `(coordinator, candidate_name, client_contact)` before creating
- **Context management** — `LIMIT 50` on the active loops SQL query + 20K char budget on the formatter (truncates with count)
- **LINK_THREAD guardrails** — three new validation checks:
  1. Target loop must exist in active loops
  2. LLM-extracted candidate/client must match target loop's actual candidate/client
  3. Target loop's candidate last name must appear in the email text (ground-truth vs ground-truth)
- **Monitoring** — Sentry measurements for `active_loops_count`, tags for guardrail retries and LINK_THREAD demotions

Also rejects `CREATE_LOOP` with "Unknown Candidate" name (feeds into error-driven retry to re-extract).

## Files changed

| File | Change |
|------|--------|
| `migrations/0013_cleanup_duplicate_loops.py` | New — marks duplicate NEW loops as cold |
| `queries/scheduling.sql` | New dedup query; LATERAL join + LIMIT 50 on active loops |
| `api/scheduling/service.py` | Dedup guard in `create_loop()` |
| `api/classifier/models.py` | `LinkThreadData` now requires `candidate_name`, `client_company` |
| `api/classifier/resolvers.py` | Rejects Unknown Candidate in `CreateLoopResolver` |
| `api/classifier/formatters.py` | `char_budget` param on `format_active_loops()` |
| `api/classifier/loop_classifier.py` | Semantic LINK_THREAD guardrails + Sentry monitoring |
| `tests/test_classifier_hook.py` | Updated test helpers for new action_data shape |
| `tests/test_classifier_resolvers.py` | Updated to test Unknown Candidate rejection |

## Test plan

- [x] All 310 existing tests pass (`pytest` — 2 pre-existing failures unrelated to this PR)
- [ ] Run migration `0013` against staging DB — verify `SELECT state, count(*) FROM loops GROUP BY state` shows reduced `new` count
- [ ] Verify classifier still works end-to-end via add-on test coordinator
- [ ] Monitor Sentry for `classifier.active_loops_count` and `classifier.guardrail_retry` tags

### DIFF
--- a/services/api/queries/scheduling.sql
+++ b/services/api/queries/scheduling.sql
@@ -101,6 +101,18 @@ LIMIT 10;
 -- Loops
 -- ============================================================
 
+-- name: find_active_loop_by_candidate_name_at_client^
+-- Check if an active loop already exists for this candidate-at-client combo.
+SELECT l.id
+FROM loops l
+JOIN candidates cand ON cand.id = l.candidate_id
+WHERE l.coordinator_id = :coordinator_id
+  AND LOWER(TRIM(cand.name)) = LOWER(TRIM(:candidate_name))
+  AND (l.client_contact_id = :client_contact_id
+       OR (l.client_contact_id IS NULL AND :client_contact_id IS NULL))
+  AND l.state NOT IN ('complete', 'cold')
+LIMIT 1;
+
 -- name: create_loop^
 INSERT INTO loops (id, coordinator_id, client_contact_id, recruiter_id, client_manager_id, candidate_id, title, state, notes)
 VALUES (:id, :coordinator_id, :client_contact_id, :recruiter_id, :client_manager_id, :candidate_id, :title, :state, :notes)
@@ -206,6 +218,8 @@ ORDER BY l.updated_at DESC;
 
 -- name: get_active_loops_full_for_coordinator
 -- Active loops (not complete or cold) for a coordinator, with actors.
+-- Sorted by most recent loop_event (true activity signal), falling back
+-- to loop.created_at for loops with no events yet.
 SELECT
     l.id                AS loop_id,
     l.coordinator_id    AS loop_coordinator_id,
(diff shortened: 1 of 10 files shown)


services/api/queries/scheduling.sql
```diff
@@ -101,6 +101,18 @@ LIMIT 10;
...
 -- ============================================================
 
+-- name: find_active_loop_by_candidate_name_at_client^
+-- Check if an active loop already exists for this candidate-at-client combo.
+SELECT l.id
+FROM loops l
+JOIN candidates cand ON cand.id = l.candidate_id
+WHERE l.coordinator_id = :coordinator_id
+  AND LOWER(TRIM(cand.name)) = LOWER(TRIM(:candidate_name))
+  AND (l.client_contact_id = :client_contact_id
```
nsadeh: I think it's better to use the client company name rather than a contact ID.

services/api/src/api/classifier/loop_classifier.py
```diff
@@ -293,4 +339,84 @@ def _apply_guardrails(
                 f"or use CREATE_LOOP",
             )
 
+        # 6. CREATE_LOOP must have a real candidate name
+        if item.action == SuggestedAction.CREATE_LOOP:
+            cand_name = (item.action_data.get("candidate_name") or "").strip()
+            if not cand_name or cand_name.lower() == "unknown candidate":
+                return (
+                    item.model_copy(update={"action": SuggestedAction.NO_ACTION}),
```
nsadeh, line 350: How does this result in being re-submitted to the classifer?

services/api/queries/scheduling.sql
nsadeh: Good call — switched the dedup query, migration, and `create_loop()` to match on `LOWER(TRIM(cc.company))` via a LEFT JOIN on `client_contacts` instead of comparing `client_contact_id` directly. This way different contacts at the same firm still dedup correctly.

Changed in: `scheduling.sql` (query + migration), `service.py` (new `client_company` param), `resolvers.py` and `addon/routes.py` (pass `client_company` through).

services/api/src/api/classifier/loop_classifier.py
nsadeh: The retry path is in `classify()` (lines 180–205): `_apply_guardrails()` returns `(item, error_string)`. The error strings are collected into `guardrail_errors`. When **all** suggestions fail guardrails (`guardrail_errors and not valid_items`), the errors are joined and injected into the LLM prompt via `retry_input = context_input.model_copy(update={"error": error_msg})`, then `classify_new_thread` is called again. The LLM sees the error message (e.g. "candidate_name must be a real name, not 'Unknown Candidate'") and gets a second chance to extract the correct name.

Added a clarifying comment above the check to make this flow more obvious.

PR #80, opened 2026-05-13
feat(agent): XML I/O, conversation-history retries, expire_suggestion

nsadeh:
## Summary

Aligns the `next-action-agent` Python pipeline with the rewritten LangFuse prompt:

- **Input:** `NextActionInput` slimmed to four XML-formatted fields — `date`, `thread_history`, `email`, `loops`. New formatters render emails (`<email direction='inbound|outbound'>...`) and loops (`<loop id='lop_…'><stage>…</stage><actors>…</actors><pending-suggestions>…</pending-suggestions></loop>`). XML escaping on every interpolated value.
- **Output parsing:** extracts JSON array from `<suggestions>[…]</suggestions>` envelope. Tolerates markdown fences, surrounding chatter, and missing tags.
- **Retries via conversation history:** the input-field `error` / `coordinator_response` indirection is gone. On batch-guardrail violations or parse failures the agent appends `assistant` + `user(follow-up)` turns and retries once with the same input. For coordinator responses, the worker now fetches the originating `ASK_COORDINATOR` suggestion and the agent splices a reconstructed prior assistant turn + the coordinator's answer into the conversation.
- **New `expire_suggestion` action:** schema + per-item guardrail (target must be a known pending id) + auto-resolver + `SuggestionService.expire_by_id` + aiosql query.
- **Multi-loop guardrails:** recruiter draft cap (≤ distinct known recruiters on the thread) and single-client-draft cap. Both surface as retryable errors with the user-specified messages.

## Files

- Schema / models: `classifier/{schemas,models}.py`
- Formatters: `classifier/formatters.py` (additive; existing formatters retained for loop classifier)
- Agent: `classifier/next_action_agent.py` (bypasses the generic `LLMEndpoint`; drives `llm.complete()` directly for conversation history)
- Resolver + service: `classifier/{resolvers,service}.py`
- SQL: `queries/suggestions.sql` (new `expire_suggestion` query, no-op when row isn't pending)
- Worker: `gmail/workers.py` (`process_coordinator_response` plumbs `originating_suggestion`)
- Tests: `tests/test_classifier_hook.py` (existing tests rewritten to mock the new pipeline; new coverage for XML formatters, batch guardrails, conversation retry, coordinator response splicing, and expire_suggestion)

## Test plan

- [x] `cd services/api && uv run pytest tests/test_classifier_hook.py tests/test_classifier_formatters.py tests/test_classifier_resolvers.py tests/test_gmail_hooks.py tests/test_gmail_webhook.py -q` → 92/92 pass
- [x] `cd services/api && uv run ruff check src/api/classifier/ src/api/gmail/workers.py tests/test_classifier_hook.py` → clean
- [ ] Local end-to-end: trigger a multi-loop two-recruiter thread, confirm batch guardrail retries land cleanly in LangFuse trace
- [ ] Resolve an `ASK_COORDINATOR` via the addon → new LangFuse trace shows 4-message conversation with reconstructed assistant turn
- [ ] Confirm `expire_suggestion` flips the target row to `expired` in `agent_suggestions`

## Notes for reviewer

- `services/api/tests/test_ai_endpoint.py::TestLLMEndpointCall::test_successful_call_returns_parsed_output` is failing on this branch **and on `main`** — unrelated to this PR (spec mock missing `.name`). Flagged separately.
- The CLAUDE.md "Multi-loop draft strategy" memory remains accurate; the new guardrails enforce what was previously aspirational.
- No new env vars, no Railway changes — references/ docs not touched.

### DIFF
--- a/services/api/src/api/addon/routes.py
+++ b/services/api/src/api/addon/routes.py
@@ -919,10 +919,10 @@ async def _handle_recruiter_selected(body: AddonRequest, svc: LoopService, email
     Three callers, distinguished by which extra params come along:
 
     - **UPDATE_ACTOR path** (``update_actor_role`` present): the coordinator
-      picked from the autocomplete on an UPDATE_ACTOR card. Delegate to
-      ``_handle_update_actor`` — autocomplete-selection IS the commit for
-      this card. (For mid-type values that don't parse as "Name <email>",
-      ``_handle_update_actor`` no-ops and refreshes.)
+      picked from the autocomplete on an UPDATE_ACTOR card. STAGE the pick
+      on the suggestion's ``action_data.pending_pick`` and refresh — the
+      card re-renders with the inputs pre-filled, and the coordinator must
+      click Save to actually commit. Autocomplete selection never commits.
 
     - **JIT path** (``draft_id`` present): the coordinator picked a recruiter
       from the autocomplete on a DRAFT_EMAIL card. Stash on draft pending
@@ -941,10 +941,13 @@ async def _handle_recruiter_selected(body: AddonRequest, svc: LoopService, email
     def _field(name: str) -> str | None:
         return _get_form_value(body, name)
 
-    # UPDATE_ACTOR origin — checked FIRST so we don't fall through to the
-    # create-loop form rebuild (which silently wipes the overview).
+    # UPDATE_ACTOR origin — stage the pick rather than commit. Misclicks in
+    # the autocomplete dropdown were committing directly to the loop with no
+    # way to undo; this routes through the standard Save-button commit step.
     if suggestion_id and update_actor_role:
-        return await _handle_update_actor(body, svc, email, **kwargs)
+        return await _stash_update_actor_pick(
+            body, svc, email, request, suggestion_id, update_actor_role
+        )
 
     if draft_id and suggestion_id:
         # JIT path — stash the pick on draft.pending_jit_data instead of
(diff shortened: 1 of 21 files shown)


services/api/src/api/classifier/formatters.py
```diff
@@ -225,6 +226,161 @@ def format_pending_suggestions(suggestions: list[Suggestion]) -> str:
...
+    cc = loop.client_contact
+    if cc is None:
+        return "<client-contact>Unknown</client-contact>"
+    name = cc.name or cc.email
+    company = cc.company or ""
+    label = f"{name}, {company}".rstrip(", ").strip()
+    return f"<client-contact>{_escape(label or 'Unknown')}</client-contact>"
+
+
+def _format_pending_suggestions_xml(pending: list[Suggestion]) -> str:
```
nsadeh, line 367: let's make this a bit richer no? Format each pending suggestion type's action_data well? There shouldn't be many, so the context shouldn't bloat. We just need to present draft_email and ask_coordinator, since the others are auto-resolved so they'll never be pending.

nsadeh: Done in 2f4cf06. The pending block now renders action_data per type: `draft_email` gets `<summary>`, `<recipient-type>`, `<body>`; `ask_coordinator` gets `<summary>` and `<question>`. Other action types fall back to summary-only, but per your point those don't end up pending in practice (auto-resolved). Test in `TestXMLFormatters.test_loop_xml_renders_actors_and_pending` covers both shapes.

services/api/src/api/addon/routes.py
```diff
@@ -1455,8 +1455,17 @@ async def _handle_accept_suggestion(body: AddonRequest, svc: LoopService, email:
     return await _build_refreshed_overview(request, email)
 
 
+# Reject re-runs the agent for these action types — DRAFT_EMAIL and
+# ASK_COORDINATOR are the cases where a different next attempt is plausible.
+# ADVANCE_STAGE / NO_ACTION / EXPIRE_SUGGESTION rejections stay dead-ends.
+_RERUN_ON_REJECT_ACTIONS = frozenset({SuggestedAction.DRAFT_EMAIL, SuggestedAction.ASK_COORDINATOR})
```
nsadeh: These can just be any suggestion that isn't auto-resolved. The other ones are not dead ends, they are just always resolved. The coordinator has no mechanism to reject them.

services/api/src/api/classifier/router.py
```diff
@@ -65,28 +65,36 @@ async def on_email(
...
+        if linked_loops:
+            # Linked thread → Next Action Agent (inbound and outgoing).
+            # Internal-only messages are kept here: recruiter→coordinator on a
+            # live loop is signal, not noise.
+            await self._agent.act(event, linked_loops, arq_pool=arq_pool)
+            return
+
+        # 3. Unlinked thread. Now the internal-only filter applies — random
+        # LRP-to-LRP chatter shouldn't try to spawn a new scheduling loop.
         if _is_internal_only(msg):
```
nsadeh: Is there _any_ way to run this check on the thread participants rather than just message participants?

Scenario:
1. Client requests interview (pre-agent, or agent made a mistake)
2. Coordinator forwards to recruiter by hand
3. Recruiter responds

There should be a loop created on 3, no?

services/api/src/api/addon/routes.py
nsadeh: Good call. Done in d305e27 — pulled an `AGENT_AUTO_RESOLVE_ACTIONS` constant out of `resolvers.py` (frozen from `build_agent_registry().keys()` at import time) and flipped the route check to `suggestion.action not in AGENT_AUTO_RESOLVE_ACTIONS`. Functionally the same set today, but now anyone adding a new auto-resolver doesn't have to remember to update the route.

services/api/src/api/classifier/router.py
nsadeh: Yes, fixed in d305e27. `_is_internal_only` now walks `event.thread_messages` instead of just looking at the trigger message — your exact scenario (external client → coord forwards → recruiter replies) is covered by `test_forwarded_client_request_then_internal_reply_classifies`. The recruiter's all-LRP reply still triggers `classifier.classify()` because the client is upstream in the thread. Filter only fires when the *entire* thread is LRP-to-LRP.

services/api/src/api/classifier/next_action_agent.py
```diff
@@ -135,22 +114,73 @@ async def act(
...
+                input=context_input.model_dump(),
+            ):
+                self._langfuse.update_current_span(
+                    metadata={
+                        "prompt_name": "next-action-agent",
+                        "prompt_version": prompt.version,
+                        "prompt_labels": prompt.labels,
+                    }
+                )
+                messages = self._build_initial_messages(
```
nsadeh, line 160: Can we show this as the input instead? I want to be able to see the actual input messages. Or is that not how it's supposed to work? I feel like we're missing resonses to feedback and errors.

nsadeh: Right call — that was hiding the actual conversation. Fixed in f808d26.

`_call_llm` now calls `update_current_span(input=messages)` right before dispatching to the LLM service. Retries route through this function with the appended assistant + user-feedback messages, so the final span input is the full conversation including:
- the system prompt + initial user message (the XML input)
- any prior `<suggestions>` the model emitted (assistant turn)
- the feedback message we sent back ("Your previous response could not be parsed. Error: …" / "The coordinator responded with…" / rejection follow-up)

The original `NextActionInput` dataclass moves to `metadata.context` so the structured input is still inspectable when you need to cross-check what we built, but it doesn't pretend to be what the LLM saw.

PR #87, opened 2026-05-14
feat(classifier): rewire loop classifier to v26 XML prompt + add CM/client guardrails

nsadeh:
## Summary

Wires the `LoopClassifier` caller to the v26 XML prompt format (the prompt was rewritten v22→v26 to mirror what PR #80 did for `next-action-agent`) and adds two new CREATE_LOOP guardrails the prompt asked for.

The runtime now bypasses the generic `llm_endpoint` and drives `llm.complete()` directly — same pattern as `NextActionAgent` — so it can parse the `<suggestions>[...]</suggestions>` envelope the prompt emits and retry via conversation history when guardrails fail. The deterministic eval against v26 went from 10/17 pass with ~8 JSON-fix retries to **11/17 pass with 0 JSON-fix retries** — the prompt's reasoning preamble was fighting the old endpoint's "respond ONLY with JSON" injection, and removing that conflict halves the loop classifier's LLM round-trips on a typical run.

## What changed

**New shared module** [`agent_runtime.py`](services/api/src/api/classifier/agent_runtime.py)
- `parse_suggestions_envelope`, `diagnostics_from_response`, `build_error_followup`, `SuggestionsParseError`. Both `LoopClassifier` and `NextActionAgent` now use these — the duplication that was sitting in `next_action_agent.py` is gone.

**Three new sibling XML formatters** in `formatters.py`
- `format_email_xml_simple` — `<email direction='inbound|outbound'><timestamp>…</timestamp><from>Name <email></from>…</email>` (addresses with angle brackets, XML-escaped). Matches v26's leaner shape.
- `format_thread_history_xml_simple` — concatenated `<email>` blocks, oldest-first, per-message direction derived from `coordinator_email`, with the same truncation marker pattern as `format_thread_history_xml`.
- `format_active_loops_xml` — `<loop id="…"><client-company>…</client-company><candidate>…</candidate></loop>` blocks. Candidate name + client company only — the rich loop XML used by `next-action-agent` is too much for thread-to-loop matching.
- The PR #80 formatters (`format_email_xml` etc.) stay untouched; they continue to serve `next-action-agent`.

**Schema** ([`schemas.py`](services/api/src/api/classifier/schemas.py))
- `LoopClassifierInput`: `active_loops_summary` → `active_loops` (matches v26's template variable name), `error` field removed (retries ride conversation history now, not an input field).

**LoopClassifier rewrite** ([`loop_classifier.py`](services/api/src/api/classifier/loop_classifier.py))
- Mirrors `NextActionAgent`'s `act()` / `_call_llm` / `_classify_with_messages` structure. Span input is the live messages list, so retry follow-ups show up in LangFuse traces. Per-call diagnostics (finish_reason, completion_tokens, model, provider, latency) attach to the span on success and to `LoopClassifierError` on failure.

**Endpoint removed**
- `classify_new_thread` deleted from [`endpoints.py`](services/api/src/api/classifier/endpoints.py) — no callers left after the rewire. Tests updated to patch `fetch_prompt` + `_llm.complete` instead.

**Two new CREATE_LOOP guardrails**
- **CM allow-list:** `cm_email` (lowercased) must be one of the canonical 12 client managers from the prompt's `<client-managers>` block. Case-insensitive so the model's `RQuatroni → rquatroni` normalization habit no longer trips it.
- **Client contact validity:** `client_name` must not equal `candidate_name` (case-insensitive). `client_email`, when present, must not be `@longridgepartners.com` and must not be a consumer-email domain (gmail/yahoo/hotmail/outlook/icloud/aol/proton/hey/qq/163/naver/etc — 26 domains). When `client_name` / `client_email` are absent, the loop is accepted — automated ATS notifications (Greenhouse) often arrive without an extractable named client contact, and we want loops to flow rather than block.
- Guardrails return `(item, error)` without coercing `item.action` to `NO_ACTION` on failure. The caller drops the item on `error is not None` and the conversation-history retry feeds the error back to the model with a follow-up like *"cm_email 'someone-else@…' is not on the permitted client-manager list. The CM MUST be one of: adam@…, alowe@…, …"* — so the model gets to self-correct rather than the runtime silently demoting.

## Test plan

- [x] `cd services/api && uv run pytest tests/test_classifier_hook.py tests/test_classifier_formatters.py -q` → 91/91 pass (8 new guardrail tests + the existing surface)
- [x] `cd services/api && uv run ruff check src/api/classifier/ tests/test_classifier_hook.py` → clean
- [x] `cd services/api && uv run python scripts/eval_loop_classifier_xml.py` → 11/17 pass, 0 JSON-fix retries against v26 (vs 10/17 with 8 retries before). The remaining 5 failures are all model-side normalization (CM email case, `Balyasny Asset Management L.P.` long-form, model picking a different client_name on one Citadel item) — none caused by format mismatch anymore.
- [ ] Local end-to-end: drop a real inbound on the dev API, confirm the LangFuse trace's prompt input matches v26's example shape (no `Name email` plain style — addresses are now `Name <email>` with XML-escaped brackets; active loops are the leaner `<loop>` form). Trigger a guardrail-failure path and confirm the messages list in the trace ends with `assistant` + `user(follow_up)` turns.

## Reviewer notes

- The classifier no longer has a JSON-fix retry path via the LLMEndpoint. If the model emits something un-parseable, the conversation-history retry in `_classify_with_messages` covers it — `SuggestionsParseError` triggers a retry with a follow-up instructing the model to produce a valid envelope.
- Consumer-email blocklist lives as a module-level frozenset in `loop_classifier.py` next to the CM allow-list. If we hit false positives in prod traces, extend the frozenset and ship. If the list ever needs to be tunable per-coordinator, that's when we should extract it.
- No env vars added, no Railway changes, no migrations — `references/` docs not touched.


services/api/src/api/classifier/agent_runtime.py
```diff
@@ -0,0 +1,103 @@
...
+SUGGESTIONS_RE = re.compile(r"<suggestions>(.*?)</suggestions>", re.DOTALL)
+# Fallback: first bare JSON array we can find.
+JSON_ARRAY_RE = re.compile(r"\[[\s\S]*\]")
+
+
+class SuggestionsParseError(Exception):
+    """Raised when an LLM response can't be parsed into a SuggestionItem list."""
+
+
+def parse_suggestions_envelope(content: str) -> ClassificationResult:
```
nsadeh, line 58: Does this need to be duplicated

services/api/src/api/classifier/agent_runtime.py
```diff
@@ -0,0 +1,103 @@
...
+    for raw in data:
+        try:
+            suggestions.append(SuggestionItem.model_validate(raw))
+        except ValidationError as exc:
+            raise SuggestionsParseError(f"suggestion item failed schema validation: {exc}") from exc
+
+    return ClassificationResult(suggestions=suggestions)
+
+
+def diagnostics_from_response(response: LLMResponse, *, attempt: int) -> dict:
```
nsadeh: Does this need to be duplicated? CAn we bundle this into an update for the old LLM client?

services/api/src/api/classifier/agent_runtime.py
nsadeh: Kept it (now documented in the module docstring as of e9c4984). The shape differs from `LLMEndpoint._try_parse` in two ways:

1. **Contract.** This parses `reasoning preamble + <suggestions>[…]</suggestions>` (what the v26 loop-classifier and #80 next-action-agent prompts emit). `_try_parse` parses a bare JSON object against a caller-supplied Pydantic schema (what `extract_create_loop_fields` and friends emit).
2. **Output type.** This is hard-coded to `ClassificationResult` / `SuggestionItem`. Unifying would mean either making the endpoint classifier-domain-aware (bad layering) or breaking the extractor's "any caller-supplied schema" contract.

Both agents call the same function — no duplication between LoopClassifier and NextActionAgent. The only overlap with `_try_parse` is the markdown-fence stripping, which is too small to extract.

services/api/src/api/classifier/agent_runtime.py
nsadeh: Done — moved to `LLMResponse.to_diagnostics(*, attempt)` as a method on the dataclass in e9c4984. Pure derivation from response fields, naturally belongs on the response object. Updated both call sites (`next_action_agent.py` and `loop_classifier.py`); the test stub now constructs a real `LLMResponse` instead of a `MagicMock` so the method actually runs in tests rather than silently returning a child mock.

rfcs/rfc-gmail-workspace-addon.md
```diff
@@ -68,6 +68,7 @@ The sidebar UI is built from **Google's Card v2 framework** — a declarative JS
...
+  - **Stale-homepage caveat & current handling.** After a coordinator acts on a suggestion inside a contextual (thread) card, Gmail may keep serving its cached homepage card on next view — showing the just-resolved suggestion. Two mitigations are in place: (1) **all mutating `/addon/action` handlers are idempotent** — they no-op (returning the scoped post-action view) if the suggestion/draft is no longer `pending`/`generated`, so a stale double-click can never re-send an email or re-advance a stage; (2) mutating handlers return the `renderActions` envelope with a **top-level `stateChanged: true`** sibling, which per Google's docs lets Gmail clear cached state. Two earlier `stateChanged` placements were reverted (`6d878dc`: bare-`action` sibling; `0b995bd`: nested inside `renderActions.action`) — both rejected by Gmail's validator. The current top-level-sibling placement is the documented one; whether Gmail honors it for the *homepage* surface specifically is verified empirically (manual test) — and the idempotency guard means the bug is harmless even if it doesn't. The homepage `↻ Refresh` button (`OpenLink` overlay + `onClose:RELOAD`) remains the reliable manual remedy.
```
nsadeh: Ok but this has been resolved now!

services/api/src/api/addon/contact_inputs.py
```diff
@@ -22,27 +25,19 @@
 
 def build_recruiter_inputs(
```
nsadeh, line 26: I think the extra email input now is unneeded right?

services/api/src/api/addon/routes.py
```diff
@@ -1466,6 +1504,15 @@ async def _handle_discard_draft(body: AddonRequest, svc: LoopService, email: str
         return await _build_refreshed_overview(request, email)
 
     draft = await draft_svc.get_draft(draft_id)
+    # Idempotency guard: stale double-click must not re-discard / re-reject
+    # an already-finalized draft. Only GENERATED is discardable.
+    if draft and draft.status != DraftStatus.GENERATED:
```
nsadeh: Why would the idempotency guard be here?

It should be pretty simple:

A suggestion can not be acted upon unless it is pending.

Am I missing something?

services/api/src/api/addon/routes.py
```diff
@@ -1546,6 +1615,17 @@ async def _handle_reject_suggestion(body: AddonRequest, svc: LoopService, email:
     suggestion_svc = SuggestionService(db_pool=svc._pool)
     suggestion = await suggestion_svc.get_suggestion(suggestion_id)
 
+    # Idempotency guard: stale double-click on an already-resolved
+    # suggestion must not re-discard a draft or re-enqueue the rejection
+    # re-run.
+    if suggestion is None or suggestion.status != SuggestionStatus.PENDING:
```
nsadeh: Seems like this needs to be refactored into a shared guardrail function?

services/api/src/api/classifier/loop_classifier.py
```diff
@@ -230,32 +351,188 @@ async def classify(
...
+                )
+            raise LoopClassifierError(
+                str(exc),
+                raw_responses=responses,
+                call_diagnostics=diagnostics,
+            ) from exc
+
+        # Run each suggestion through guardrails — collect valid ones and
+        # the human-readable per-item errors that the retry follow-up needs.
+        valid_items: list[SuggestionItem] = []
```
nsadeh, line 427: This logic seems to be repeated in multiple places?

rfcs/rfc-gmail-workspace-addon.md
nsadeh: Resolved in 81b7a8a — the RFC note (line 71) now reads as confirmed-resolved rather than a hedged caveat: the renderActions + top-level stateChanged:true envelope is the primary, prod-verified fix (it clears the cached homepage); the idempotency guard and the manual Refresh button are reframed as defense-in-depth, not the remedy.

services/api/src/api/addon/contact_inputs.py
nsadeh: Right — collapsed to a single directory-autocomplete input in 81b7a8a. The autocomplete emits one "Name <email>" token and every commit handler already does parse_name_email(raw_name) or parse_name_email(raw_email), so the second field was dead UI. build_recruiter_inputs now returns one widget (dropped email_field/prefill_email); updated the three overview/cards.py callers, the required-gate, and the tests.

services/api/src/api/addon/routes.py
nsadeh: Agreed — centralized in 81b7a8a. There's now one helper, _stale_action_guard(request, email, body, *, suggestion=None, draft=None), placed right after _post_mutation_view. It returns None when the suggestion is still PENDING (or draft GENERATED) and otherwise returns the scoped post-action view. Each mutating handler calls it in one line right after fetching its suggestion/draft, before any side effect — so 'a suggestion cannot be acted upon unless it is pending' is now enforced in exactly one place.

services/api/src/api/addon/routes.py
nsadeh: Done in 81b7a8a — extracted _stale_action_guard and collapsed the six duplicated guard blocks (accept / reject / respond_to_question / update_actor / send_draft / discard_draft) into one-line calls to it. _handle_create_loop keeps its own bespoke check since that's a different concern (dedupe loop creation; the manual path has no suggestion to guard).

services/api/src/api/classifier/loop_classifier.py
nsadeh: You're right that the parse → guardrail → conversation-history-retry → diagnostics loop appears in three direct-LLM agents (LoopClassifier._classify_with_messages, NextActionAgent._act_with_messages, create_loop_extractor._extract_with_messages). I deliberately did NOT unify it in this PR: the three differ in their middle step — LoopClassifier runs suggestion guardrails, NextActionAgent validates the <suggestions> envelope, the extractor parses a single <loop> object — so a shared runner would need a strategy/callback seam, and all three paths are currently prod-confirmed. The shared primitives (envelope parsers, build_error_followup, LLMResponse.to_diagnostics) are already factored into agent_runtime; folding the orchestration loop itself together is a worthwhile but higher-risk follow-up I'd rather do on its own PR with its own test pass, not bundled into this rewire. Filing it as a follow-up.

PR #93, opened 2026-05-19
fix(agent): stop client CC leaking onto recruiter drafts

nsadeh:
## Summary

A recruiter-bound draft (`recipient_type=recruiter`) with the recruiter slot unset (JIT pending) was leaking a **client-domain address onto the CC line**. Root cause: `_is_forward_draft` returned `False` ("reply") whenever `to_emails` was empty, which fired reply-all CC carry-over while the recruiter's identity was still unknown — dragging the trigger thread's client CC onto the draft.

## What changed

- **Rule 1 — `_strip_external_cc`** (`services/api/src/api/drafts/service.py`): for `recipient_type in ("recruiter", "internal")`, the CC list keeps only addresses at `INTERNAL_DOMAIN` (reused from `api.classifier.router`). Inverts the problem — don't hunt client addresses, allow only internal. Wired as a backstop *after* `_apply_reply_all_cc`.
- **Rule 2 — `_is_forward_draft`**: empty `to_emails` now returns `True` (forward) → reply-all carry-over skipped → CM-only CC until the coordinator fills the contact. The `recipient_type == "client"` privacy guard is still checked first.

## Reviewer notes

- Test fixtures in `test_draft_service.py` had `lrp.com` as a synthetic internal stand-in; swapped to `longridgepartners.com` so the real `INTERNAL_DOMAIN` constant matches. External domains untouched.
- Added 4 tests: screenshot-bug repro (unset recruiter + client on trigger CC), recruiter-reply client-strip, and two `_is_forward_draft` Rule-2 units.
- No new env vars, no Railway changes.

## Test plan

- [x] `ruff check src/ tests/` + `ruff format --check` — clean
- [x] `pytest tests/test_draft_service.py` — 43 passed
- [ ] Adjacent suites (`test_addon_send_draft`, `test_draft_cards`, `test_addon`, `test_scheduling_integration`, `test_overview_service`) — recommend CI confirm (local runs kept auto-backgrounding)

### DIFF
--- a/services/api/src/api/drafts/service.py
+++ b/services/api/src/api/drafts/service.py
@@ -212,32 +212,15 @@ def _apply_reply_all_cc(
     return result
 
 
-def _strip_external_cc(
-    cc_emails: list[str],
-    recipient_type: RecipientType | None,
-) -> list[str]:
-    """Keep only internal (LRP-domain) addresses on recruiter/internal CC.
+def _internal_only(emails: list[str]) -> list[str]:
+    """Drop every non-internal (non-LRP-domain) address from ``emails``.
 
-    Recruiter-bound drafts and internal notes are LRP-internal coordination
-    the client must never see. Reply-all carry-over
-    (``_apply_reply_all_cc``) can drag *any* address off the trigger
-    thread's CC line onto ours — including client-side participants who were
-    CC'd on the upstream request (e.g. a client teammate who is not even
-    ``loop.client_contact``). Rather than trying to identify who is
-    "client-side", we invert it: on these comms, allow only the internal
-    domain through. Anything else is dropped.
-
-    Mirror of the client privacy guard in ``_is_forward_draft`` — same
-    intent (no internal/external leakage), opposite direction.
-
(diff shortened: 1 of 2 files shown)


services/api/src/api/drafts/service.py
```diff
@@ -202,6 +212,35 @@ def _apply_reply_all_cc(
     return result
 
 
+def _strip_external_cc(
```
nsadeh: This is confusing code. This function _may_ strip external CC's _if_ the recipient type is a certain type. In addition, the CC part is unnecessary, since we're passign a list of emails - this could be used anywhere on any list of emails.

Better ot have it _always_ strip externals from the list that is passed in, and let the call site handle the logic of whether to call it.

nsadeh: Done in a89758e. Replaced with pure `_internal_only(emails)` — always strips non-internal from whatever list it's handed, no recipient_type logic inside and not CC-specific. The recipient_type gate now lives at the `resolve_reply_recipients` call site.

PR #100, opened 2026-06-01
feat(agent): schedule_interview suggestion type — bones only

nsadeh:
## Summary

Adds the data path for booked interviews end-to-end without any side-effecting integrations (no calendar invites, no conflict checks). The agent can now propose \`schedule\`, \`update\`, and \`cancel\` ops against a \`scheduled_interviews\` table; coordinators approve from the sidebar.

## What changed

- **\`ScheduleInterviewData\`** pydantic model with per-op (schedule/update/cancel) null-rule validators; registered in \`ACTION_DATA_MODELS\` so guardrails bounce malformed payloads through the existing retry-with-explanation loop.
- **\`scheduled_interviews\`** table (migration 0015) + aiosql queries + \`ScheduledInterviewService\` for CRUD against it.
- **Sidebar builder** renders Schedule / Update / Cancel cards with the candidate + client company in the header; update ops diff old → new per changed field. Accept left, dismiss right per existing convention.
- **NextActionAgent input context** now includes a \`<scheduled-interviews>\` block per loop (active only — canceled rows are hidden from the LLM).
- **Accept handler** dispatches by \`interview_op\` to create / update / cancel the row, then resolves the suggestion as \`ACCEPTED\`. Reject path reuses the generic rejection branch since \`SCHEDULE_INTERVIEW\` is intentionally not auto-resolved.

## Why now

This is the "bones" only — wires the data flow end-to-end so we can iterate on richer behavior (calendar invites, conflict checking, attendee management) without rebuilding the plumbing. Coordinator approval is on the critical path for every interview op, matching the UX philosophy in CLAUDE.md.

## Test plan

- [ ] \`yoyo apply\` migrations against local Postgres; \`\\d scheduled_interviews\` shows expected schema
- [ ] Craft an agent response with \`interview_op=schedule\` and \`interview_start_time=null\` → confirm one error-followup round
- [ ] Seed a \`schedule_interview\` suggestion → sidebar header reads \`Schedule interview for <candidate> with <client_company>\`, time renders ET, Accept on left
- [ ] Accept schedule → row in \`scheduled_interviews\`, suggestion ACCEPTED
- [ ] Accept update changing only time → diff shows \`Time: <old> → <new>\` and nothing else
- [ ] Accept cancel → \`is_canceled = true\`
- [ ] Re-trigger NextActionAgent on a thread with an active interview → \`<scheduled-interviews>\` block present in rendered XML
- [ ] Dismiss a pending \`schedule_interview\` → REJECTED + \`process_rejection_response\` enqueued, no DB mutation to \`scheduled_interviews\`

### DIFF
--- a/services/api/queries/scheduling.sql
+++ b/services/api/queries/scheduling.sql
@@ -411,10 +411,12 @@ WHERE loop_id = :loop_id AND is_canceled = false
 ORDER BY interview_start_time;
 
 -- name: update_scheduled_interview!
+-- PATCH semantics: any bound param that is NULL means "leave the existing value
+-- alone." Lets the agent send only the fields it wants to change.
 UPDATE scheduled_interviews
-SET interview_start_time = :interview_start_time,
-    interview_duration   = :interview_duration,
-    interview_notes      = :interview_notes,
+SET interview_start_time = COALESCE(:interview_start_time, interview_start_time),
+    interview_duration   = COALESCE(:interview_duration, interview_duration),
+    interview_notes      = COALESCE(:interview_notes, interview_notes),
     updated_at           = now()
 WHERE id = :id;
 
--- a/services/api/queries/scheduling.sql
+++ b/services/api/queries/scheduling.sql
@@ -383,3 +383,43 @@ SELECT id, loop_id, gmail_thread_id, subject, linked_at
 FROM loop_email_threads
 WHERE loop_id = ANY(:loop_ids)
 ORDER BY loop_id, linked_at;
+
+-- ============================================================
+-- Scheduled interviews
+-- ============================================================
+
+-- name: create_scheduled_interview^
+INSERT INTO scheduled_interviews (
+    id, loop_id, interview_start_time, interview_duration, interview_notes
+)
+VALUES (:id, :loop_id, :interview_start_time, :interview_duration, :interview_notes)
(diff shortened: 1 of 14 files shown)


services/api/queries/scheduling.sql
```diff
@@ -383,3 +383,43 @@ SELECT id, loop_id, gmail_thread_id, subject, linked_at
...
+-- Non-canceled interviews for a loop, earliest first.
+SELECT id, loop_id, interview_start_time, interview_duration, interview_notes,
+       is_canceled, created_at, updated_at
+FROM scheduled_interviews
+WHERE loop_id = :loop_id AND is_canceled = false
+ORDER BY interview_start_time;
+
+-- name: update_scheduled_interview!
+UPDATE scheduled_interviews
+SET interview_start_time = :interview_start_time,
```
nsadeh: Should these be coalesce, in case the update statement/classifier returns null for fields it doesn't want to update?

nsadeh (commented):
Just the one nit

services/api/queries/scheduling.sql
nsadeh: Done in f2e42fb — switched to PATCH semantics. The UPDATE now COALESCEs each field against its current value, and I relaxed the validator on `update` so only `interview_id` is required (start_time / duration / notes can all be null = "leave alone"). The card builder distinguishes `null` (omitted) from `""` (explicit clear).

